PR #11, opened 2024-03-11
Add internal_mac and dedicated_server_credential data source

hollow:
This pull request adds two features to the provider:

- A field `internal_mac` for `dedicated_server` to be able to use the MAC address of the internal interface in `post_install_script` of `leaseweb_dedicated_server_installation`
- A data source to fetch credentials, such as those for REMOTE_MANAGEMENT to store them in 1Password via Terraform


nrocco (approved):
Looks good to me

PR #14, opened 2024-07-19
Implement instances & new terraform framework

alrayyes:


nrocco (approved):

README.md
majidkarimizadeh, line 1: Can you pls add the architecture design in readme https://medium.com/@matiasvarela/hexagonal-architecture-in-go-cfd4e436faa3?

internal/provider/data_sources/public_cloud/model/auto_scaling_group.go
majidkarimizadeh, line 1: As we talked, it would be better if we can merge `data_source` and `resources` models together

internal/handlers/public_cloud/handler.go
majidkarimizadeh, line 1: `handler` and `convertor`! these are generic words, can we rename them so we can know what they're doing by their name?

examples/README.md
majidkarimizadeh, line 1: afaik the examples are generating docs by terraform, can we add more example so the users will be able to use it better?

majidkarimizadeh (commented):
Well done @alrayyes some small items need to be addressed, and we're good to go.

internal/core/domain/ips.go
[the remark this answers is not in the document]
alrayyes: Yes. Instances is a list of instances and nothing else. In this case there is no extra logic, but there is extra logic in other collections. This also helps with code completion in the ide and static analysis, as we know that this can only contain Instances and nothing else.

internal/core/domain/instance.go
```diff
@@ -0,0 +1,161 @@
...
+	instance.SshKey = optional.SshKey
+
+	if optional.RootDiskSize != nil {
+		instance.RootDiskSize = *optional.RootDiskSize
+	}
+
+	return instance
+}
+
+func NewUpdateInstance(
```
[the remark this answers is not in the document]
alrayyes: Those are all the variables that are supported  by the api for updating instances. As this is business logic I created 2 seperate constructor functions (NewUpdateInstance/NewCreationInstance). The whole point of domain options is that they are passed around instead of options.

This is however confusing. In hindsight this is more the responsibility of the handler. I will update the createInstance/updateInstance functions there and remove them from the domain object.

internal/core/domain/instance.go
```diff
@@ -0,0 +1,161 @@
...
+	instance.MarketAppId = optional.MarketAppId
+	instance.SshKey = optional.SshKey
+	instance.StartedAt = optional.StartedAt
+	instance.PrivateNetwork = optional.PrivateNetwork
+	instance.AutoScalingGroup = optional.AutoScalingGroup
+
+	return instance
+}
+
+func NewCreateInstance(
```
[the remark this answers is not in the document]
alrayyes: This is going bye bye.

internal/core/ports/public_cloud_repository.go
```diff
@@ -0,0 +1,62 @@
...
+	) (*domain.Instance, *shared.RepositoryError)
+
+	// UpdateInstance updates an instance in the public cloud api.
+	UpdateInstance(
+		instance domain.Instance,
+		ctx context.Context,
+	) (*domain.Instance, *shared.RepositoryError)
+
+	// DeleteInstance deletes an instance in the public cloud api.
+	DeleteInstance(id value_object.Uuid, ctx context.Context) *shared.RepositoryError
```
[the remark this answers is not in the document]
alrayyes: Nope. We only care about the id here. This keeps things nice and simple. A domain object should be passed if deletion is dependent on some business logic. It's the same as GetInstance, there's no point in passing  a whole domain object.

internal/core/ports/public_cloud_service.go
```diff
@@ -0,0 +1,47 @@
...
+	) (*domain.Instance, *shared.ServiceError)
+
+	// UpdateInstance updates an instance.
+	UpdateInstance(
+		instance domain.Instance,
+		ctx context.Context,
+	) (*domain.Instance, *shared.ServiceError)
+
+	// DeleteInstance deletes an instance.
+	DeleteInstance(id value_object.Uuid, ctx context.Context) *shared.ServiceError
```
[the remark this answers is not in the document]
alrayyes: See above :)

README.md
alrayyes: Will do! I'll save it till last though as it isn't necessary for functionality.

internal/provider/data_sources/public_cloud/model/auto_scaling_group.go
alrayyes: I will try!

internal/handlers/public_cloud/handler.go
alrayyes: Suggestions are welcome.

examples/README.md
alrayyes: I don't know if multiple examples are supported (see [current examples](https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/docs/resources/dedicated_server.md) as well). I've only used required options in the examples on purpose. Those can be copy pasted for any use case. The rest of the generated documentation is pretty clear as to what optional values are supported. That shouldn't be an issue for people that already use terraform.

internal/core/domain/instance.go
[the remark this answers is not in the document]
majidkarimizadeh: then let's remove it

internal/core/domain/instance.go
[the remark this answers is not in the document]
majidkarimizadeh: then let's remove it

internal/core/ports/public_cloud_repository.go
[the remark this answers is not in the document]
majidkarimizadeh: ok so let's keep it like this for now

internal/provider/data_sources/public_cloud/model/auto_scaling_group.go
majidkarimizadeh: since it's not a blocker, let's work on it after release, I'll create an issue.

internal/handlers/public_cloud/handler.go
majidkarimizadeh: `convertor -> dataAdapter`
`handler -> facade`
directory: `handlers -> facades`

majidkarimizadeh (approved):

PR #45, opened 2024-08-26
feat: implement dedicated_server control panels datasource

arunvarghe:
- Add script to have control panels data source
- Unit test
- Acceptance test


Caddyfile
```diff
@@ -8,4 +8,4 @@ route /publicCloud/v1/* {
 route /bareMetals/v2/* {
 	uri strip_prefix /bareMetals/v2
 	reverse_proxy prism_dedicatedServer:4010
-}
```
alrayyes, line 11: An accidental commit, we've all been there. Please revert this as it has nothing to do with the PR

examples/data-sources/leaseweb_control_panels/data-source.tf
```diff
@@ -0,0 +1,2 @@
+# List all Control panels
+data "leaseweb_control_panels" "all" {}
```
alrayyes, line 2: Are there any other control panels? If there are the data source name needs to be more specific.

internal/core/domain/dedicated_server/control_panel.go
```diff
@@ -0,0 +1,19 @@
...
+	Id   string
+	Name string
+}
+
+// NewControlPanel creates a new controlPanel with all supported values.
+func NewControlPanel(
+	id string,
+	name string,
+) ControlPanel {
+	controlPanel := ControlPanel{
```
alrayyes, line 13: This can be replaces with

```go
return ControlPanel{
Id: id,
Name name,
}
```

internal/core/domain/dedicated_server/control_panel_test.go
```diff
@@ -0,0 +1,16 @@
+package dedicated_server
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func Test_NewControlPanel(t *testing.T) {
+	controlPanel := NewControlPanel(
```
alrayyes, line 9: It's better to use the want/got pattern here:

```go
got := NewControlPanel("id", "name")
want := dedicated_server.ControlPanel(Id: "id", Name: "name")

assert.Equal(t, want, got)
```

internal/core/domain/dedicated_server/control_panel_test.go
```diff
@@ -0,0 +1,16 @@
+package dedicated_server
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func Test_NewControlPanel(t *testing.T) {
```
alrayyes, line 8: When it comes to tests "public" function tests should be camelCase. So in this case `TestNewControlPanel`.
Only private functions should have an underscore. So `newControlPanel` would be `Test_newControlPanel`

internal/core/domain/dedicated_server/control_panels.go
```diff
@@ -0,0 +1,3 @@
+package dedicated_server
+
+type ControlPanels []ControlPanel
```
alrayyes, line 3: Nice

internal/core/ports/dedicated_server_repository.go
```diff
@@ -14,4 +14,10 @@ type DedicatedServerRepository interface {
 		domain.DedicatedServers,
 		*shared.RepositoryError,
 	)
+
+	// GetAllControlPanels Retrieve all control panels from the dedicated server api.
```
alrayyes, line 18: This should be a sentence as documentation is generated from the comment. So:

`// GetAllControlPanels retrieves all control panels from the dedicated_server api.`

Also try to keep naming consistent. If we use `dedicated_server` we should it everywhere.

internal/core/ports/dedicated_server_service.go
```diff
@@ -11,4 +11,7 @@ import (
 type DedicatedServerService interface {
 	// GetAllDedicatedServers gets all dedicated servers.
 	GetAllDedicatedServers(ctx context.Context) (*domain.DedicatedServers, *errors.ServiceError)
+
+	// GetAllControlPanels gets all control panels.
```
alrayyes, line 15: Here you got it right :-)

alrayyes: No need for the pointer return here. `GetAllDedicatedServers(ctx context.Context) (domain.DedicatedServers, *errors.ServiceError)` with also accept `return nil, nil`

alrayyes: I believe the following is also valid:

```go
dedicatedServers := DedicatedServers{}
if dedicatedServers == nil {
return true
}
// Returns true
```

internal/core/services/dedicated_server/service.go
```diff
@@ -29,6 +29,22 @@ func (srv Service) GetAllDedicatedServers(ctx context.Context) (
...
+
+	controlPanels, err := srv.dedicatedServerRepository.GetAllControlPanels(ctx)
+	if err != nil {
+		return nil, errors.NewFromRepositoryError(
+			"GetAllControlPanels",
+			*err,
+		)
+	}
+
+	return &controlPanels, nil
```
alrayyes, line 45: `return controlPanels, nil` should also be ok.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -28,6 +30,13 @@ func (r *repositorySpy) GetAllDedicatedServers(ctx context.Context) (
...
 }
 
+func (r *repositorySpy) GetAllControlPanels(ctx context.Context) (
+	domain.ControlPanels,
+	*sharedRepository.RepositoryError,
+) {
+	return r.controlPanels, r.getAllControlPanelError
+}
+
 func newRepositorySpy() repositorySpy {
```
alrayyes, line 40: There's not really a need for this in the tests, but it doesn't hurt.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -99,3 +108,70 @@ func TestService_GetAllDedicatedServers(t *testing.T) {
 		},
 	)
 }
+
+func TestService_GetAllControlPanels(t *testing.T) {
+	t.Run(
+		"service passes back control panels from repository",
+		func(t *testing.T) {
+
+			want := domain.ControlPanels{
```
alrayyes, line 117: see, want got :-)

internal/core/services/dedicated_server/service_test.go
```diff
@@ -99,3 +108,70 @@ func TestService_GetAllDedicatedServers(t *testing.T) {
...
+				domain.ControlPanel{
+					Id:   "id",
+					Name: "name",
+				},
+			}
+
+			spy := newRepositorySpy()
+			spy.controlPanels = want
```
alrayyes, line 125: See, there's no point in the `newRepositorySpy()` function. To keep tests simple you set attributes individually anyway. So you may as well do:

```go
spy := RepositorySpy{controlPanels: want}
```

This keeps the code more succinct and easier to read. I should have done this from the beginning as well.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -99,3 +108,70 @@ func TestService_GetAllDedicatedServers(t *testing.T) {
...
+			assert.Error(t, err)
+			assert.ErrorContains(t, err, "some error")
+		},
+	)
+
+	t.Run(
+		"error from repository getAllControlPanels bubbles up", func(t *testing.T) {
+			service := New(
+				&repositorySpy{
+					controlPanels: domain.ControlPanels{
```
alrayyes, line 159: I don't understand the point of this.

internal/facades/dedicated_server/data_adapters/to_data_source_model/adapter_test.go
```diff
@@ -342,3 +342,24 @@ func TestAdaptDedicatedServers(t *testing.T) {
 	assert.Equal(t, "id1", got.DedicatedServers[0].Id.ValueString())
 	assert.Equal(t, "id2", got.DedicatedServers[1].Id.ValueString())
 }
+
+func TestAdaptControlPanels(t *testing.T) {
+	id := "id"
+	name := "name"
+	controlPanels := domain.ControlPanels{
```
alrayyes, line 349: There's no need for this amount of detail as `TestAdaptControlPanel` already tests the `adaptControlPanel` function. To keep the test code to the minimal it's recommended to use mocks:

```go
got := AdaptControlPanels(domain.ControlPanels{Id: "id"}

assert.Len(t, got.ControlPanels, 1)
assert.Equal(t, id, got.ControlPanels[0].Id.ValueString())
```

This way you won't have to update this function is `NewControlPanel` ever changes.

internal/facades/dedicated_server/facade.go
```diff
@@ -29,7 +29,22 @@ func (h DedicatedServerFacade) GetAllDedicatedServers(ctx context.Context) (
 	return &dataSourceDedicatedServers, nil
 }
 
-func NewDedicatedServerFacade(dedicatedServerService ports.DedicatedServerService) DedicatedServerFacade {
+// GetAllControlPanels retrieve all operating systems.
```
alrayyes, line 32: //GetAllControlPanels retrieves all control panels.

internal/facades/dedicated_server/facade.go
```diff
@@ -15,11 +15,11 @@ type DedicatedServerFacade struct {
 }
 
 // GetAllDedicatedServers retrieve all dedicated servers.
```
alrayyes, line 17: Typo here:

`// GetAllDedicatedServers retrieves all dedicated servers.`

internal/facades/dedicated_server/data_adapters/to_data_source_model/adapter_test.go
```diff
@@ -342,3 +342,24 @@ func TestAdaptDedicatedServers(t *testing.T) {
...
+	}
+
+	got := AdaptControlPanels(controlPanels)
+
+	assert.Len(t, got.ControlPanels, 1)
+	assert.Equal(t, id, got.ControlPanels[0].Id.ValueString())
+	assert.Equal(t, name, got.ControlPanels[0].Name.ValueString())
+}
+
+func TestAdaptControlPanel(t *testing.T) {
```
alrayyes, line 360: The function name should be `Test_adaptControlPanel`

internal/provider/data_sources/dedicated_server/control_panels/schema.go
```diff
@@ -0,0 +1,30 @@
...
+				Computed: true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"id": schema.StringAttribute{
+							Computed:    true,
+							Description: "The control panel unique identifier",
+						},
+						"name": schema.StringAttribute{
+							Computed:    true,
+							Description: "The name of the control panel",
```
alrayyes, line 23: I know this is copied from the api definitions, but descriptions like this are useless. `The name of the control panel` can be inferred by anyone with more than 2 brain cells. I think it's better not to include useless documentation like this at all (for the end user). That way if the user reads the generated markdownfile (s)he knows that everything in it is important.

internal/repositories/dedicated_server_repository/data_adapters/to_domain_entity/adapter_test.go
```diff
@@ -445,3 +445,13 @@ func Test_adaptDedicatedServer(t *testing.T) {
 	)
 	assert.Equal(t, want, got)
 }
+
+func Test_adaptControlPanel(t *testing.T) {
+	controlPanel := dedicatedServer.NewControlPanel()
```
alrayyes, line 450: This is why you need to set the fields as `required` in the api definitions. This constructor function is useless as it does nothing. If the fields in de api definitions are set to `required`

```json
 "controlPanels": {
        "description": "A list of control panels",
        "items": {
          "title": "controlPanel",
          "description": "A single control panel",
          "properties": {
            "id": {
              "description": "The unique ID of this control panel",
              "type": "string"
            },
            "name": {
              "description": "A human readable name describing the control panel",
              "type": "string"
            }
          },
          "required": ["id", "name"],
          "type": "object"
        },
        "type": "array"
      }
```

 this would then be:

`controlPanel := dedicatedServer.NewControlPanel("id", "name")

internal/repositories/dedicated_server_repository/dedicated_server_repository.go
```diff
@@ -78,6 +79,50 @@ func (p DedicatedServerRepository) GetAllDedicatedServers(ctx context.Context) (
...
+	request := d.dedicatedServerApi.GetControlPanelList(d.authContext(ctx))
+
+	result, response, err := request.Execute()
+
+	if err != nil {
+		return nil, shared.NewSdkError("GetAllControlPanels", err, response)
+	}
+
+	metadata := result.GetMetadata()
+	pagination := shared.NewPagination(
```
alrayyes, line 97: Pagination, nice!

alrayyes, line 97: Might be a good idea to implement threading (see how instances are retrieved). It won't be much faster but you can learn how it works.

alrayyes (changes requested):
Good stuff, a few small issues

alrayyes:
It also appears that you haven't enabled the git hooks :-). Run the following:

`pnpm i`

The current commit message also needs to be changed to adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). In this case it should be something like `feat: implement dedicated_server control panels datasource`

internal/repositories/dedicated_server_repository/data_adapters/to_domain_entity/adapter_test.go
arunvarghe: Should it be updated in api definition?

examples/data-sources/leaseweb_control_panels/data-source.tf
imtiazPabel: Last time we had the name of `leaseweb_dedicated_server_control_panels`. Shouldn't we keep the same name for maintaining the legacy?
 https://github.com/Leaseweb/terraform-provider-leaseweb/blob/v0.4.1/examples/data-sources/leaseweb_dedicated_server_control_panels/data-source.tf#L2

internal/core/domain/dedicated_server/control_panel_test.go
imtiazPabel: @alrayyes, it's called `PascalCase(TestNewControlPanel)`. Also, where it's specified that public/private functions should have different way of test method? 

I'm asking because of, I just found, the test method should be PascalCase.

examples/data-sources/leaseweb_control_panels/data-source.tf
alrayyes: @imtiazPabel So much other stuff has changed in the provider that we may as well adopt a tabula rasa attitude towards this. Current clients are going to have to update their schemas anyway.

I do think we should write down naming rules somewhere.

internal/core/domain/dedicated_server/control_panel_test.go
alrayyes: @imtiazPabel Thank you , that is correct. `PascalCase`. That's how goland generates tests for me. I'm assuming that that's the correct way. It also makes sense. Otherwise you will get nonsensical tests function names like `TestdoSomething` `TestcalculateSomething`.

The underscore rule is only for "private" functions.

examples/data-sources/leaseweb_control_panels/data-source.tf
imtiazPabel: Well, I feel, it should be discussed within the team. Because of, I'm on the same phase. I maintain my data source as `leaseweb_dedicated_server_operating_systems`. In that case, one of us, should change the data source name to be on the same page.

alrayyes: We should all by using the same naming schema, I agree. Longer strings are annoying for the end user, but inconsistency is more annoying.

internal/repositories/dedicated_server_repository/data_adapters/to_domain_entity/adapter_test.go
alrayyes: yes

internal/core/domain/dedicated_server/control_panel_test.go
imtiazPabel: @alrayyes, PascalCase also make sense to me. I'm wondering about the underscore case for the private function. It would be appreciated, instead of assuming, if you could be confirmed.

https://go.dev/doc/tutorial/add-a-test
https://quii.gitbook.io/learn-go-with-tests/go-fundamentals/select

Everywhere I looked, I only found PascalCase. While they may not provide an example for private functions, they also didn't mention that private function names should use underscore case.

I don't trust goland, because of, goland also suggest the package name, which isn't recommended by community.

alrayyes: Indeed, all of the examples given are for exported functions.

This isn't [jetbrains opinion](https://youtrack.jetbrains.com/issue/GO-5185):

> In my opinion, this is the expected behaviour for unexported functions. Anyway, IDE uses [go tests](https://github.com/cweill/gotests) for tests generation, so it can be changed only there.

internal/core/ports/dedicated_server_repository.go
imtiazPabel: @alrayyes, which document is generated from the comment?

alrayyes: https://pkg.go.dev/github.com/leaseweb/terraform-provider-leaseweb@v1.3.2-alpha#section-documentation.

You also see the documentation & examples when you hover over functions in the IDE.

internal/repositories/dedicated_server_repository/dedicated_server_repository.go
arunvarghe: as discussed, not required

internal/core/services/dedicated_server/service_test.go
arunvarghe: actually it is not required

Caddyfile
alrayyes, line 1: The file is still in the commit.  But let's ignore this for now.

internal/facades/dedicated_server/data_adapters/to_data_source_model/adapter_test.go
```diff
@@ -342,3 +342,18 @@ func TestAdaptDedicatedServers(t *testing.T) {
 	assert.Equal(t, "id1", got.DedicatedServers[0].Id.ValueString())
 	assert.Equal(t, "id2", got.DedicatedServers[1].Id.ValueString())
 }
+
+func TestAdaptControlPanels(t *testing.T) {
+	got := AdaptControlPanels(domain.ControlPanels{domain.NewControlPanel("id", "name")})
+
+	assert.Len(t, got.ControlPanels, 1)
+	assert.Equal(t, "id", got.ControlPanels[0].Id.ValueString())
+	assert.Equal(t, "name", got.ControlPanels[0].Name.ValueString())
```
alrayyes, line 351: There's no need to validate all fields, that's already done in `Test_adaptControlPanel`. To make this test more concise you could do something like:

```go
got := AdaptControlPanels(domain.ControlPanels{Id: "id"}}

assert.Len(t, got.ControlPanels, 1)
assert.Equal(t, "id", got.ControlPanels[0].Id.ValueString())
```

That way you also don't have to update this test every time `NewControlPanel` changes.

internal/facades/dedicated_server/facade_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (s *serviceSpy) GetAllControlPanels(ctx context.Context) (*domain.ControlPanels, *serviceErrors.ServiceError) {
+	return &s.controlPanels, s.getAllControlPanelError
+}
+
+func newServiceSpy() serviceSpy {
+	return serviceSpy{}
+}
+
+func TestFacade_GetAllControlPanels(t *testing.T) {
```
alrayyes, line 40: Since `GetAllControlPanels` is an exported function, the test name should be `TestGetAllControlPanels`. If you right click on a function in goland you can generate tests automatically.

internal/provider/data_sources/dedicated_server/control_panels/read.go
```diff
@@ -0,0 +1,40 @@
...
+	ctx context.Context,
+	_ datasource.ReadRequest,
+	resp *datasource.ReadResponse,
+) {
+
+	tflog.Info(ctx, "Read control panels")
+	controlPanels, err := c.client.DedicatedServerFacade.GetAllControlPanels(ctx)
+
+	if err != nil {
+		resp.Diagnostics.AddError("Unable to read control panels", err.Error())
```
alrayyes, line 21: We should discuss this, but it makes sense for this to be `Unable to read dedicated_server_control_panels`, no?

internal/repositories/dedicated_server_repository/dedicated_server_repository.go
```diff
@@ -78,6 +79,50 @@ func (p DedicatedServerRepository) GetAllDedicatedServers(ctx context.Context) (
...
+		request,
+	)
+
+	for {
+		result, response, err := request.Execute()
+		if err != nil {
+			return nil, shared.NewSdkError("GetAllControlPanels", err, response)
+		}
+
+		for _, sdkControlPanel := range result.GetControlPanels() {
```
alrayyes, line 109: It would make more sense here to implement an `AdaptControlPanels` function, which loops through the sdk control panels & calls `adaptControlPanel`.

alrayyes (changes requested):

alrayyes:
Looks good! Just one thing about the commits, they should be in present tense as it's easier to read. This will become important for the auto generated changelog & release notes in the future.

So `feat: fixed comments about test cases and documentation` should be `feat: fix comments about test cases and documentation`

alrayyes:
In commit messages `feat` should only be used for introducing new features to the code base. So it should be 'chore(docs): fix comments about test cases and documentation` or `docs: fix comments about test cases and documentation`. 

I recommend taking 10 minutes to go through the [conventional commits summary](https://www.conventionalcommits.org/en/v1.0.0/#summary). The reason i'm nitpicking this is that eventually we want to auto generate changelogs & release notes from the commit messages. This only works if the conventional commit standard is properly adhered to.

internal/core/domain/dedicated_server/control_panel.go
```diff
@@ -0,0 +1,17 @@
+package dedicated_server
+
+type ControlPanel struct {
+	Id   string
+	Name string
+}
+
+// NewControlPanel creates a new controlPanel with all supported values.
```
alrayyes, line 8: `// NewControlPanel creates ControlPanel.` if you want to be pedantic here and clickable links.

alrayyes (changes requested):

alrayyes (approved):

PR #50, opened 2024-08-27
Docs/update docs

alrayyes:
Update documentation for launching/updating instances

PR #54, opened 2024-08-28
[WIP] feat: add import dedicated server

majidkarimizadeh:


internal/core/ports/dedicated_server_repository.go
```diff
@@ -26,4 +26,10 @@ type DedicatedServerRepository interface {
 		domain.ControlPanels,
 		*shared.RepositoryError,
 	)
+
+	// GetDedicatedServer return dedicated_server from the dedicated server api.
```
alrayyes, line 30: `returns data for a singular dedicated_server from the dedicated server api.` would be better here.

internal/core/ports/dedicated_server_service.go
```diff
@@ -17,4 +17,7 @@ type DedicatedServerService interface {
 
 	// GetAllControlPanels gets dedicated_server.ControlPanels.
 	GetAllControlPanels(ctx context.Context) (domain.ControlPanels, *errors.ServiceError)
+
+	// GetDedicatedServer return dedicated server.
```
alrayyes, line 21: See ports comment.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -183,3 +192,62 @@ func TestService_GetAllControlPanels(t *testing.T) {
 		},
 	)
 }
+
+func TestService_GetDedicatedServer(t *testing.T) {
```
alrayyes, line 196: `TestServiceGetDedicatedServer`

internal/core/services/dedicated_server/service_test.go
```diff
@@ -183,3 +192,62 @@ func TestService_GetAllControlPanels(t *testing.T) {
...
+	t.Run(
+		"service passes back dedicated server from repository",
+		func(t *testing.T) {
+
```
alrayyes, line 200: For tests it's best to use the `arrange / act / assert` pattern and group that way. An example of this would be:

```go
id := "tralala"
want := domain.DedicatedServer{Id: id}
spy := repositorySpy{dedicatedServer: want}

got, err := service.GetDedicatedServer(context.TODO(), id)

assert.Nil(t, err)
assert.Equal(t, want, *got)
````

This way the developer knows exactly what belongs where in every test. TLDR: Lose the extra spaces :p

File: internal/core/services/dedicated_server/service_test.go [hunk 5]
```diff
@@ -183,3 +192,62 @@ func TestService_GetAllControlPanels(t *testing.T) {
...
+			)
+
+			_, err := service.GetDedicatedServer(context.TODO(), "id")
+
+			assert.Error(t, err)
+			assert.ErrorContains(t, err, "some error")
+		},
+	)
+
+	t.Run(
```
alrayyes, line 235: I don't immediately see the point of this test (help me out, I'm in a different timezone).

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter.go
```diff
@@ -0,0 +1,204 @@
+package to_resource_model
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	domain "github.com/leaseweb/terraform-provider-leaseweb/internal/core/domain/dedicated_server"
+	resourcesModel "github.com/leaseweb/terraform-provider-leaseweb/internal/provider/resources/dedicated_server/model"
+)
+
+func AdaptDedicatedServer(dedicatedServer domain.DedicatedServer) resourcesModel.DedicatedServer {
```
alrayyes, line 9: Good idea to add a comment here as well as it's an exported function

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,289 @@
...
+	domain "github.com/leaseweb/terraform-provider-leaseweb/internal/core/domain/dedicated_server"
+	resourceModel "github.com/leaseweb/terraform-provider-leaseweb/internal/provider/resources/dedicated_server/model"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_adaptRack(t *testing.T) {
+	rack := domain.NewRack("id", "cap", "type")
+
+	got := adaptRack(rack)
+	want := resourceModel.Rack{
```
alrayyes, line 16: Nice

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,289 @@
...
+	pciCard := domain.NewPciCard("description")
+	got := adaptPciCard(pciCard)
+	want := resourceModel.PciCard{
+		Description: basetypes.NewStringValue("description"),
+	}
+	assert.Equal(t, want, got)
+}
+
+func Test_adaptPciCards(t *testing.T) {
+	pciCard1 := domain.NewPciCard("description1")
```
alrayyes, line 183: It's better to use a mock here so this test doesn't have to be refactored every time `NewPciCard` changes. So I would make this something like `pciCards := {{description: "description1"}, {description: "description2"}}`

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,289 @@
...
+	want := resourceModel.PciCards{
+		resourceModel.PciCard{Description: basetypes.NewStringValue("description1")},
+		resourceModel.PciCard{Description: basetypes.NewStringValue("description2")},
+	}
+	assert.Len(t, got, 2)
+	assert.Equal(t, want, got)
+}
+
+func Test_adaptHdds(t *testing.T) {
+	hdds := domain.Hdds{
```
alrayyes, line 196: Shorthand can be used here:

```go
hdds := domain.Hdds{{Id: "id1"}, {Id: "id2"}}
```

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,289 @@
...
+	assert.Equal(t, want, got)
+}
+
+func Test_adaptSpecs(t *testing.T) {
+	specs := domain.NewSpecs(
+		"chassis",
+		true,
+		domain.Cpu{Quantity: 1},
+		domain.Ram{Size: 2},
+		domain.Hdds{domain.Hdd{Id: "id"}},
```
alrayyes, line 226: shorthand can be used here as well.

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func TestAdaptDedicatedServer(t *testing.T) {
+
+	specs := domain.NewSpecs(
+		"chassis",
+		true,
+		domain.Cpu{Quantity: 1},
+		domain.Ram{Size: 2},
+		domain.Hdds{domain.Hdd{Id: "id1"}},
```
alrayyes, line 245: Shorthand :)

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,289 @@
...
+		domain.PciCards{domain.NewPciCard("d")},
+	)
+	featureAvailability := domain.FeatureAvailability{IpmiReboot: false}
```
alrayyes, line 248: As these variables aren't used anymore it makes sense to just use the values themselves in the `New` function below, no?

internal/facades/dedicated_server/facade.go
```diff
@@ -61,8 +66,25 @@ func (f DedicatedServerFacade) GetAllControlPanels(ctx context.Context) (
 	return &dataSourceControlPanels, nil
 }
 
+// GetDedicatedServer returns dedicated server details.
+func (f DedicatedServerFacade) GetDedicatedServer(ctx context.Context, id string) (
+	*resourceModel.DedicatedServer,
+	*shared.FacadeError,
+) {
+	dedicatedServer, err := f.dedicatedServerService.GetDedicatedServer(ctx, id)
```
alrayyes, line 74: You can call `GetDedicatedServer` in the `GetDedicatedServers` function now. Advantage is that you can refactor `TestService_GetDedicatedServers` to make the test simpler as you've already tested edge cases in `TestService_GetDedicatedServer`.

internal/provider/dedicated_server_resource_test.go
```diff
@@ -0,0 +1,10 @@
+package provider
+
+import (
+	"testing"
+)
+
+// TODO: Implement acceptance tests for import dedicated server.
+func TestAccDedicatedServerResource(t *testing.T) {
```
alrayyes, line 8: This is easy to do. See the public_cloud resource acceptance tests. It should be handled automagically. You can also check out the acceptance test part of the custom terraform provider tutorial.

internal/provider/resources/dedicated_server/dedicated_server/configure.go
```diff
@@ -0,0 +1,35 @@
...
+	resp *resource.ConfigureResponse,
+) {
+	if req.ProviderData == nil {
+		return
+	}
+
+	coreClient, ok := req.ProviderData.(client.Client)
+
+	if !ok {
+		resp.Diagnostics.AddError(
```
alrayyes, line 23: Discussed this with @imtiazPabel. We should refactor this at some point as the code is repeated over and over again.

internal/provider/resources/dedicated_server/dedicated_server/resource.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/provider/client"
+)
+
+var (
+	_ resource.Resource                = &dedicatedServerResource{}
+	_ resource.ResourceWithConfigure   = &dedicatedServerResource{}
+	_ resource.ResourceWithImportState = &dedicatedServerResource{}
+	_ resource.ResourceWithModifyPlan  = &dedicatedServerResource{}
```
alrayyes, line 14: Remove this interface implementation if you're not using it. The same for the other ones so we can get rid of the panics below.

internal/provider/resources/dedicated_server/dedicated_server/schema.go
```diff
@@ -0,0 +1,239 @@
...
+			},
+			"null_routed": schema.BoolAttribute{
+				Computed: true,
+			},
+			"ports": portSchema(),
+		},
+	}
+}
+
+func portSchema() schema.ListNestedAttribute {
```
alrayyes, line 225: nice

internal/provider/resources/dedicated_server/model/dedicated_server.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+type DedicatedServer struct {
+	Id                  types.String        `tfsdk:"id"`
+	AssetId             types.String        `tfsdk:"asset_id"`
+	SerialNumber        types.String        `tfsdk:"serial_number"`
+	Rack                Rack                `tfsdk:"rack"`
+	Location            Location            `tfsdk:"location"`
```
alrayyes, line 12: Are you sure this works?

alrayyes (commented):
Looks good! I still highly recommend setting expected fields to `required` in the api definitions so that they are correct there.

internal/core/services/dedicated_server/service_test.go
majidkarimizadeh: why!?
https://github.com/Leaseweb/terraform-provider-leaseweb/blob/4bafe06b199790e212aac2d0da21e1bc936d4a14/internal/core/services/public_cloud/service_test.go#L282

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
majidkarimizadeh: making it long line, it is hard to read, I'd not do that in this case

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
majidkarimizadeh: same, hard to read

internal/provider/dedicated_server_resource_test.go
majidkarimizadeh: That's a good one @alrayyes, because I guess it's using the same variable in create to do the import and since we don't have create in dedicated server it's making it a bit tricky.

I checked the public cloud one:
https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/internal/provider/instance_resource_test.go#L86

there is no straightforward way to do the import with the dedicated server ID and checking the Terraform state.
pls correct me if I'm wrong.

internal/core/services/dedicated_server/service_test.go
majidkarimizadeh: you're right <3

internal/provider/resources/dedicated_server/dedicated_server/resource.go
majidkarimizadeh: I can get rid of only`ResourceWithModifyPlan`, for create/update/delete come with the `Resource` interface. pls correct me if I'm wrong.

internal/core/services/dedicated_server/service_test.go
majidkarimizadeh: liked it <3

internal/facades/dedicated_server/facade.go
majidkarimizadeh: in `GetDedicatedServers` we get the batch dedicated servers but in `GetDedicatedServer` we use ID to get one single resource, other than one function return `resourceModel` and the other one return `dataModel` so it seems merging these 2 making it more complicated! what do you think!?

internal/provider/resources/dedicated_server/model/dedicated_server.go
majidkarimizadeh: Basically we're still using the duplicated `dataModel` here in the resource directory, still we didn't add any custom login which make a differentiated between `resourceModel` and `dataModel`! also, we are looking into merging these 2 models if it's applicable! since we don't have tests for `dataModels` is it a good idea to add tests here? or wait for the result of merging investigation?!

internal/provider/resources/dedicated_server/dedicated_server/configure.go
imtiazPabel: Basically, the Configure method is completely/almost identical regardless of every data source/ resources. We discussed that it would be better if we could write it once and use it everywhere.

internal/core/services/dedicated_server/service_test.go
imtiazPabel: @alrayyes, through `IDE and go test` generator the method name comes same as @majidkarimizadeh defined

imtiazPabel:
I'm missing the directory `examples/resources/leaseweb_dedicated_server/`

docs/resources/dedicated_servers.md
```diff
@@ -0,0 +1,222 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "leaseweb_dedicated_servers Resource - leaseweb"
```
imtiazPabel, line 3: I feel, the resource should be singular, the existing `public_cloud_instnace` is also a singular. However, the legacy was also `leaseweb_dedicated_server`

docs/resources/dedicated_servers.md
```diff
@@ -0,0 +1,222 @@
...
+- `port` (String)
+
+
+<a id="nestedatt--private_networks"></a>
+### Nested Schema for `private_networks`
+
+Read-Only:
+
+- `id` (String)
+- `link_speed` (Number)
```
imtiazPabel, line 158: Since link_speed is an enum, should be better to show the options.
https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/docs/resources/public_cloud_instance.md?plain=1#L74

An example: https://github.com/Leaseweb/terraform-provider-leaseweb/pull/50/files#diff-0db8933c108623e8a8857dccb481e5450918ed9a8395b6196a84213de80a58d2R169

docs/resources/dedicated_servers.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+<a id="nestedatt--rack"></a>
+### Nested Schema for `rack`
+
+Read-Only:
+
+- `capacity` (String)
+- `id` (String)
+- `type` (String)
```
imtiazPabel, line 171: https://github.com/Leaseweb/terraform-provider-leaseweb/pull/54/files#r1743360309

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	assert.Equal(t, want, got)
+}
+
+func Test_adaptPorts(t *testing.T) {
+	port1 := domain.NewPort("name1", "1111")
+	port2 := domain.NewPort("name2", "2222")
+	ports := domain.Ports{port1, port2}
+	got := adaptPorts(ports)
+	want := resourceModel.Ports{
```
imtiazPabel, line 79: Since `resourceModel.Ports` is slice of `resourceModel.Port`. Should be enough

```
resourceModel.Ports{
		{Name: basetypes.NewStringValue("name1"), Port: basetypes.NewStringValue("1111")},
		{Name: basetypes.NewStringValue("name2"), Port: basetypes.NewStringValue("2222")},
	}
```

It's not a deal breaker. You can consider it as a comment 😊

internal/facades/dedicated_server/data_adapters/to_resource_model/adapter_test.go
```diff
@@ -0,0 +1,275 @@
...
+		Status:    basetypes.NewStringValue("status"),
+		Subnet:    basetypes.NewStringValue("subnet"),
+		VlanId:    basetypes.NewStringValue("vlanid"),
+		LinkSpeed: basetypes.NewInt32Value(12),
+	}
+	assert.Equal(t, want, got)
+}
+
+func Test_adaptPrivateNetworks(t *testing.T) {
+	privateNetwork := domain.PrivateNetworks{
```
imtiazPabel, line 111: https://github.com/Leaseweb/terraform-provider-leaseweb/pull/54/files#r1743406380

internal/provider/resources/dedicated_server/dedicated_server/import_state.go
```diff
@@ -0,0 +1,43 @@
...
+	resource.ImportStatePassthroughID(
+		ctx,
+		path.Root("id"),
+		req,
+		resp,
+	)
+
+	dedicatedServer, err := d.client.DedicatedServerFacade.GetDedicatedServer(ctx, req.ID)
+	if err != nil {
+		resp.Diagnostics.AddError("Error importing Dedicated server", err.Error())
```
imtiazPabel, line 25: Adding the server ID in the message would be nice. So that, if there was, multiple schema for importing, customer can know for which server id import failed.

internal/provider/resources/dedicated_server/dedicated_server/schema.go
```diff
@@ -0,0 +1,239 @@
...
+			},
+			"power_ports": portSchema(),
+			"private_networks": schema.ListNestedAttribute{
+				Computed: true,
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"id": schema.StringAttribute{
+							Computed: true,
+						},
+						"link_speed": schema.Int32Attribute{
```
imtiazPabel, line 106: Since In `api-defination`, the field is an `enum`. Would be good to validate.

internal/provider/resources/dedicated_server/model/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package model implements data models for dedicated_server data sources.
+package model
```
imtiazPabel, line 2: Is it useful for any case?

imtiazPabel:
If you are leaving more commits in the PR, I would say not to make `force` anymore. It makes tough to review, to see the specific changes. If you want to merge with a single commit, we can always `Squash & Merge`, in that case.

internal/core/services/dedicated_server/service_test.go
alrayyes: Then it's good, nvm!

internal/provider/dedicated_server_resource_test.go
alrayyes: I think the idea is that the import throws no errors in the tests. As this is all the tutorial does I'm assuming that it's recommended practice. I think you can [check the state](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests), but you'll have to check the documentation. For now just running the import in acceptance tests should be fine.

internal/provider/resources/dedicated_server/dedicated_server/configure.go
alrayyes: But that's for another PR, this is fine.

internal/provider/resources/dedicated_server/dedicated_server/resource.go
alrayyes: That is correct.

internal/provider/resources/dedicated_server/model/dedicated_server.go
alrayyes: When you test this manually does it work? I don't think it should.

If above works then you don't need tests. Models are just lists of attributes with no extra logic. The only tested in resource models is `getAttributeTypes` (or whatever it's called). This method is used for object type metadata (see how public cloud models that implement other models are used).

docs/resources/dedicated_servers.md
alrayyes: We discussed this. For end users this is trivial to change. I say let's do it properly and not keep old implementation just for the convenience of 3 people, while annoying all new users of terraform.

docs/resources/dedicated_servers.md
alrayyes: Good catch @imtiazPabel! That is indeed the idea. Give the user as much information as possible beforehand. For stuff like this dynamic generation is even better (see examples).

internal/provider/resources/dedicated_server/dedicated_server/import_state.go
alrayyes: Let's write these rules down in the wiki somewhere.

internal/provider/resources/dedicated_server/model/doc.go
alrayyes: Put this in `doc.go` because the package has multiple files. This way the developer always  knows to edit `go.doc` if there are multiple files and package level documentation needs to be added.

docs/resources/dedicated_servers.md
imtiazPabel: In terms of properly, I feel that, if we are dealing singular dedicated server then it should singular the resource name as well.

docs/resources/dedicated_servers.md
majidkarimizadeh: This PR is about importing dedicated server! is this really matter in this PR!? there is no option to show to the customers, they can just import it!

internal/provider/resources/dedicated_server/dedicated_server/schema.go
majidkarimizadeh: validate for import!? we don't get any input for this from user!

internal/provider/resources/dedicated_server/model/dedicated_server.go
majidkarimizadeh: that's what I get:

```sh
terraform import leaseweb_dedicated_server.imp 10836167

leaseweb_dedicated_server.imp: Importing from ID "10836167"...
leaseweb_dedicated_server.imp: Import prepared!
  Prepared leaseweb_dedicated_server for import
leaseweb_dedicated_server.imp: Refreshing state... [id=10836167]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.
```

---

```sh

terraform state show 'leaseweb_dedicated_server.imp'
# leaseweb_dedicated_server.imp:
resource "leaseweb_dedicated_server" "imp" {
    asset_id             = "10836167"
    contract             = {
...

```

majidkarimizadeh:
> If you are leaving more commits in the PR, I would say not to make `force` anymore. It makes tough to review, to see the specific changes. If you want to merge with a single commit, we can always `Squash & Merge`, in that case.



> I'm missing the directory `examples/resources/leaseweb_dedicated_server/`

https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-documentation-generation#add-resource-import-documentation
done

majidkarimizadeh:
> I'm missing the directory `examples/resources/leaseweb_dedicated_server/`

https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-documentation-generation#add-resource-import-documentation
done

.gitignore
```diff
@@ -8,7 +8,7 @@ terraform.tfstate.backup
 *.log
 *.iml
 *.test
-
```
alrayyes, line 11: As this has nothing to do with the project it's the developers responsibility to put it in the [global .gitignore](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) file. The project .gitignore should not care or know about which IDE the developer is using.

examples/resources/leaseweb_dedicated_server/resource.tf
```diff
@@ -0,0 +1,4 @@
+# Manage example Dedicated server
+resource "leaseweb_dedicated_server" "example" {
+
```
alrayyes, line 3: Are there really no required attributes for creating a dedicated_server?

internal/provider/dedicated_server_resource_test.go
```diff
@@ -0,0 +1,24 @@
+package provider
+
+// TODO: fix the test case
```
alrayyes, line 3: tsk tsk tsk. Without tests this isn't done :p

internal/provider/resources/dedicated_server/dedicated_server/resource.go
```diff
@@ -0,0 +1,37 @@
...
+	return &dedicatedServerResource{}
+}
+
+type dedicatedServerResource struct {
+	client client.Client
+}
+
+// Create implements resource.Resource.
+func (d *dedicatedServerResource) Create(context.Context, resource.CreateRequest, *resource.CreateResponse) {
+
```
alrayyes, line 26: Create/Delete/Update do nothing. Is this intentional? If not see if you can maybe implement another interface that doesn't need these functions. If that's not possible and these should never be called then it's better to leave a panic statement in here imho.

internal/provider/resources/dedicated_server/dedicated_server/schema.go
alrayyes: Fair point. We only need to validate if user can create/update/delete.

examples/resources/leaseweb_dedicated_server/resource.tf
alrayyes: I see now, there's is only for import. In that case a resource file isn't necessary as far as i know. The import.sh should be enough.

internal/provider/resources/dedicated_server/model/control_panels.go
```diff
@@ -0,0 +1,5 @@
+package model
+
+type ControlPanels struct {
+	ControlPanels []ControlPanel `tfsdk:"control_panels"`
```
alrayyes, line 4: I think this will break once the user starts managing resources. But if it works for importing then it works!

internal/provider/resources/dedicated_server/model/dedicated_server.go
alrayyes: I'm pretty sure this will break once the user starts managing resources, but I could be wrong. But that's a problem for another day. I wouldn't merge data source & resource models just yet because of that.

internal/provider/resources/dedicated_server/model/specs.go
```diff
@@ -0,0 +1,12 @@
+package model
+
+import "github.com/hashicorp/terraform-plugin-framework/types"
+
+type Specs struct {
+	Chassis             types.String `tfsdk:"chassis"`
+	HardwareRaidCapable types.Bool   `tfsdk:"hardware_raid_capable"`
+	Cpu                 Cpu          `tfsdk:"cpu"`
+	Ram                 Ram          `tfsdk:"ram"`
+	Hdds                Hdds         `tfsdk:"hdd"`
```
alrayyes, line 10: `tfsdk:"hdds"`. Could you double check all the other stuff that should be plural?

alrayyes (changes requested):
Terraform, the tool of a 1001 implementation details.

examples/resources/leaseweb_dedicated_server/resource.tf
majidkarimizadeh: True, `import.sh` is necessary one

majidkarimizadeh:
> Create/Delete/Update do nothing. Is this intentional? If no...
> tsk tsk tsk. Without tests this isn't d....
> This is easy to do. See the public_cloud resource acceptance tests. It should...

@alrayyes since the messages on this PR is getting out of hand, let me clear the acceptance test issue.

There is no option to create the dedicated server via Leaseweb API. in this case, we can't implement create, and we just need to keep it empty. 
On the other side, we can't do much on the `import` acceptance test cases, if you check the Terraform documentations. it uses the same state on the create and tests import based on that.

BTW, I added a `TODO` to see if we can implement an acceptance test, but as I see it's not easy to implement in this situation, If you see an easy of implementing it LMK.

majidkarimizadeh:
@alrayyes btw when I'm getting 67 comments on my PR while I was focused, it means something is wrong

alrayyes:
Please, no more squashing commits while we're in the middle of a review. This is a lot of work now :p

internal/facades/dedicated_server/facade.go
alrayyes: Ah, so there is a specific endpoint for batches? In that case then `GetDedicatedServer` should not be used.

alrayyes (changes requested):
Looks good! Just a a couple of questions.

majidkarimizadeh:
new implementation #61

PR #56, opened 2024-09-06
[HALT] feat: implement dedicated server notification setting bandwidth creation

arunvarghe:


internal/core/domain/dedicated_server/notification_setting_bandwidth.go
```diff
@@ -0,0 +1,58 @@
...
+	Id                  string
+	Frequency           string
+	LastCheckedAt       string
+	Threshold           string
+	ThresholdExceededAt string
+	Unit                string
+	Actions             Actions
+}
+
+type Actions []Action
```
alrayyes, line 14: This should be in a separate file. The same goes for `Action`

internal/core/domain/dedicated_server/notification_setting_bandwidth.go
```diff
@@ -0,0 +1,58 @@
...
+		LastCheckedAt:       lastCheckedAt,
+		Threshold:           threshold,
+		ThresholdExceededAt: thresholdExceededAt,
+		Unit:                unit,
+		Actions:             actions,
+	}
+}
+
+// NewCreateNotificationSettingBandwidth creates notification setting bandwidth with only all the supported fields.
+func NewCreateNotificationSettingBandwidth(
```
alrayyes, line 44: I don't understand what the point of this is. A `New` function should always return a struct and not a pointer to a struct.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -183,3 +192,49 @@ func TestService_GetAllControlPanels(t *testing.T) {
 		},
 	)
 }
+
+func TestService_CreateNotificationSettingBandwidth(t *testing.T) {
+	t.Run(
+		"service passes back notification setting bandwidth from repository while creating a notification setting bandwidth",
```
alrayyes, line 198: imho it's best to keep these descriptions succinct. We already know which service we're testing here and that we're testing `create`. It's better for the description to be something like `passed back notificationSetting from repository`.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -183,3 +192,49 @@ func TestService_GetAllControlPanels(t *testing.T) {
...
+				ServerId:  "12345",
+				Frequency: "DAILY",
+				Threshold: "1",
+				Unit:      "Gbps",
+			}
+			spy := repositorySpy{notificationSettingBandwidth: &want}
+			service := New(&spy)
+			got, err := service.CreateNotificationSettingBandwidth(want, context.TODO())
+
+			assert.Equal(t, &want, got)
```
alrayyes, line 211: I would switch this around so it's easier to understand: `assert.Equal(t, want, *got)`.

internal/core/services/dedicated_server/service_test.go
```diff
@@ -183,3 +192,49 @@ func TestService_GetAllControlPanels(t *testing.T) {
...
+			assert.Nil(t, err)
+		},
+	)
+
+	t.Run(
+		"error from repository CreateNotificationSettingBandwidth bubbles up while creating a notification setting bandwidth",
```
alrayyes, line 217: Same here, succinct is better imho.

internal/core/shared/enum/notification_setting_bandwidth_frequency.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (f NotificationSettingBandwidthFrequency) String() string {
+	return string(f)
+}
+
+func (f NotificationSettingBandwidthFrequency) Values() []string {
+	return enum_utils.ConvertStringEnumToValues(notificationSettingBandwidthFrequencyValues)
+}
+
+const (
```
alrayyes, line 17: Nice!

internal/core/domain/dedicated_server/notification_setting_bandwidth.go
alrayyes: Now I see what you want to do here :-). it's better to implement an optional struct for optional fields. See `Instance` domain entity for an example. So the code would look something like:

```go
type Action struct {
RequiredField1 string
RequiredField2 int
OptionalField1 *string
OptionalField2 *int
}

type Optional struct {
OptionalField1 *string
OptionalField2 *int
}


func NewAction(requiredField1 string, requiredField2 int, optional Optional) Action {
action := Action{
RequiredField1: requiredField1,
RequiredField2: requiredField2,
}

if optional.OptionalField1 != nil {
action.OptionalField1 = optional.OptionalField1
}

if optional.OptionalField2 != nil {
action.OptionalField2 = optional.OptionalField2
}

return action
}

```

internal/facades/dedicated_server/data_adapters/to_domain_entity/adapter_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/provider/resources/dedicated_server/model"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestAdaptToCreateNotificationSettingBandwidthOpts(t *testing.T) {
+	t.Run("required values are set", func(t *testing.T) {
```
alrayyes, line 11: Missing test for optional fields.

internal/provider/resources/dedicated_server/notification_setting_bandwidth/delete.go
```diff
@@ -0,0 +1,13 @@
+package notification_setting_bandwidth
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (n *notificationSettingBandwidthResource) Delete(
```
alrayyes, line 8: Is delete doing nothing a feature or are you just doing this because the interface is telling you to?

internal/provider/resources/dedicated_server/notification_setting_bandwidth/read.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (n *notificationSettingBandwidthResource) Read(
+	ctx context.Context,
+	req resource.ReadRequest,
+	resp *resource.ReadResponse,
+) {
```
alrayyes, line 12: Same question as Delete

internal/provider/resources/dedicated_server/notification_setting_bandwidth/schema.go
```diff
@@ -0,0 +1,74 @@
...
+				Description: "The notification setting bandwidth unique identifier",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"frequency": schema.StringAttribute{
+				Required:    true,
+				Description: "The notification frequency. Valid options can be *DAILY* or *WEEKLY* or *MONTHLY*.",
+				Validators: []validator.String{
+					stringvalidator.OneOf(facade.GetFrequencies()...),
```
alrayyes, line 38: Nice!

internal/provider/resources/dedicated_server/notification_setting_bandwidth/update.go
```diff
@@ -0,0 +1,13 @@
+package notification_setting_bandwidth
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+)
+
+func (n *notificationSettingBandwidthResource) Update(
```
alrayyes, line 8: Same question as Delete/Read :p

internal/provider/resources/dedicated_server/notification_setting_bandwidth/validator/greater_than_zero_validator.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"strconv"
+)
+
+// GreaterThanZeroValidator ensures that the given value is greater than zero.
+type GreaterThanZeroValidator struct{}
```
alrayyes, line 11: I'm pretty sure there's already a built in terraform validator with the same functionality that you can use.

alrayyes: Oooohhh, because it's a string. Nevermind, good work!

internal/provider/resources/dedicated_server/notification_setting_bandwidth/validator/greater_than_zero_validator_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"testing"
+
+	schemaValidator "github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGreaterThanZeroValidator_ValidateString(t *testing.T) {
+	t.Run("does not set errors if the value is greater than 0", func(t *testing.T) {
```
alrayyes, line 13: I like these tests!

alrayyes (changes requested):
Good stuff! A few small comments.

internal/provider/resources/dedicated_server/notification_setting_bandwidth/delete.go
arunvarghe: Yes, right!!, As we discussed in last refinement, will do delete, read and update as different story

internal/provider/resources/dedicated_server/notification_setting_bandwidth/read.go
arunvarghe: Yes, right!!, I just did this because the interface makes it mandatory. As we discussed in the last refinement, will do delete, read and update as different story

internal/provider/resources/dedicated_server/notification_setting_bandwidth/update.go
arunvarghe: Yes, right!!, I just did this because the interface makes it mandatory. As we discussed in the last refinement, will do delete, read and update as different story

internal/provider/resources/dedicated_server/notification_setting_bandwidth/delete.go
alrayyes: Ok. I could be wrong but I'm pretty sure that there are other interfaces you can implement where `Delete` isn't mandatory. Correct me if i'm wrong.

internal/repositories/dedicated_server_repository/data_adapters/to_domain_entity/value_adapters_test.go
```diff
@@ -0,0 +1,23 @@
+package to_domain_entity
+
+import (
+	"github.com/leaseweb/leaseweb-go-sdk/dedicatedServer"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestAdaptNullableStringToValue(t *testing.T) {
```
alrayyes, line 9: I would also add an `ExampleAdaptNullableStringToValue` test here as this function is exported and meant to be re used.

alrayyes (changes requested):
Still a few things. Could you please see if there's maybe an interfaces that just expects `Create` to exist. If not the functions that need to be implemented should have `fatal` in them so they don't silently fail imho.

examples/resources/leaseweb_dedicated_server_notification_setting_bandwidth/resource.tf
```diff
@@ -0,0 +1,7 @@
+# Manage example Dedicated Server Notification Setting Bandwidth
+resource "leaseweb_dedicated_server_notification_setting_bandwidth" "example" {
+  server_id = "1234567"
+  frequency = "DAILY"
+  threshold = "1"
+  unit      = "Gbps"
```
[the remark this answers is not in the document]
alrayyes: If these are already enums please feel free to ignore my comments. Are they also set as enums in the api specifications?

PR #57, opened 2024-09-06
[HALT] feat: Add create dedicated server data traffic notification setting

majidkarimizadeh:

PR #58, opened 2024-09-09
feat: add dedicated server data traffic notification setting resource

majidkarimizadeh:
@alrayyes 
This PR is just POC, we need to discuss it further within the team and see pros/cons of it.
other implementation with the current setup #57


imtiazPabel:
Missing `examples/resource/leaseweb_dedicated_server_data_traffic_notification_setting` directory

alrayyes:
@majidkarimizadeh Is this still a poc?

internal/provider/dedicated_server_data_traffic_notification_setting_resource.go
```diff
@@ -0,0 +1,236 @@
...
+	)
+}
+
+func (d *dataTrafficNotificationSettingResource) Configure(ctx context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	if req.ProviderData == nil {
+		return
+	}
+	configuration := dedicatedServer.NewConfiguration()
+
+	// TODO: Refactor this part, ProviderData can be managed directly, not within client.
```
imtiazPabel, line 63: I'm also not happy like you about this part and changes on the `client.go` and passing the `apiKey` on the resource/data source.

I tried to make a solution for that on my PR. May be, we can have a discussion, and figure it out.

examples/resources/leaseweb_dedicated_server_data_traffic_notification_setting/resource.tf
```diff
@@ -0,0 +1,7 @@
+# Manage example Dedicated server data traffic notification
+resource "leaseweb_dedicated_server_data_traffic_notification_setting" "example" {
+  server_id = "12345"
```
imtiazPabel, line 3: On the old implementation, the key was `dedicated_server_id`
If we use `server_id`, then I'm gonna change on my side as well, to be on the same page.

internal/provider/dedicated_server_data_traffic_notification_setting_resource.go
```diff
@@ -0,0 +1,236 @@
...
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"server_id": schema.StringAttribute{
+				Required: true,
+			},
+			"frequency": schema.StringAttribute{
+				Required: true,
+				Validators: []validator.String{
+					stringvalidator.OneOf([]string{"DAILY", "WEEKLY", "MONTHLY"}...),
```
imtiazPabel, line 106: First question, is the enum list available on GO SDK? 
If yes, then I think we should reference the list, instead of specifying explicitly. In that way, if some days, list modifies, we don't have to do anything.
If no, specifying the enum properly in `api-defination`, the list can be generate

internal/provider/dedicated_server_data_traffic_notification_setting_resource.go
```diff
@@ -0,0 +1,236 @@
...
+
+	opts := dedicatedServer.NewDataTrafficNotificationSettingOpts(
+		data.Frequency.ValueString(),
+		data.Threshold.ValueString(),
+		data.Unit.ValueString(),
+	)
+	request := d.client.CreateServerDataTrafficNotificationSetting(d.authContext(ctx), data.ServerId.ValueString()).DataTrafficNotificationSettingOpts(*opts)
+	result, _, err := request.Execute()
+	if err != nil {
+		resp.Diagnostics.AddError("Error creating data traffic notification setting", err.Error())
```
imtiazPabel, line 141: Would you like to pass at least the `server id` on message, so if there was multiple resource request, customer can understand which resource is making error.

internal/provider/dedicated_server_data_traffic_notification_setting_resource.go
majidkarimizadeh: yeah true, we'll fix it

internal/provider/dedicated_server_data_traffic_notification_setting_resource.go
majidkarimizadeh: we talked about it, we'll fix it later.

imtiazPabel (approved):

PR #62, opened 2024-09-11
feat: add dedicated server(s) data source

majidkarimizadeh:


examples/data-sources/leaseweb_dedicated_server/data-source.tf
```diff
@@ -0,0 +1,4 @@
+# List all Dedicated servers
+data "leaseweb_dedicated_server" "srv" {
+  id = "12345"
```
alrayyes, line 3: Is id a string int or a uuid? If the former, strange.

examples/data-sources/leaseweb_dedicated_servers/data-source.tf
```diff
@@ -1,2 +1,10 @@
 # List all Dedicated servers
-data "leaseweb_dedicated_servers" "all" {}
+data "leaseweb_dedicated_servers" "all" {
+  reference               = "test-reference"
```
alrayyes, line 3: In the other data sources, are only required options listed in the tf file, or required & optional? If I remember correctly only required, but I could be wrong. We should be consistent here whichever way we go.

Is it possible in terraform to include multiple data-source examples? That way we could do required & optional and have the best of both worlds. Thoughts?

internal/provider/dedicated_server_data_source.go
```diff
@@ -0,0 +1,323 @@
...
+
+	apiClient := dedicatedServer.NewAPIClient(configuration)
+	d.client = apiClient.DedicatedServerAPI
+}
+
+func (d *dedicatedServerDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_dedicated_server"
+}
+
+func (d *dedicatedServerDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
```
alrayyes, line 102: I'm not a fan of putting everything in one file here. This especially makes code review more difficult than they need to be.

alrayyes: It's also more difficult for example to compare what's in the examples to what's in the schema.

alrayyes (changes requested):
I can't review this properly when everything is lumped into one file, especially on my little laptop screen :s

internal/provider/dedicated_server_data_source.go
majidkarimizadeh: @alrayyes for now we decided to continue with this format! we want to keep it simple, as much as possible!

examples/data-sources/leaseweb_dedicated_servers/data-source.tf
majidkarimizadeh: true, that was a mistake. fixed now

imtiazPabel (approved):

PR #63, opened 2024-09-11
feat: add import dedicated server

majidkarimizadeh:


internal/provider/dedicated_server_data_traffic_notification_setting_resource.go
```diff
@@ -205,7 +199,7 @@ func (d *dataTrafficNotificationSettingResource) Update(ctx context.Context, req
 		return
 	}
 
-	newData := resourceData{
+	newData := dataTrafficNotificationSettingResourceData{
```
alrayyes, line 202: Can we use a better file name here than `newData`?

internal/provider/dedicated_server_resource.go
```diff
@@ -0,0 +1,466 @@
...
+	_ resource.ResourceWithImportState = &dedicatedServerResource{}
+)
+
+type dedicatedServerResource struct {
+	// TODO: Refactor this part, apiKey shouldn't be here.
+	apiKey string
+	client dedicatedServer.DedicatedServerAPI
+}
+
+type dedicatedServerResourceData struct {
```
alrayyes, line 28: Can we split this up? Reviewing this is difficult.

alrayyes (changes requested):

PR #64, opened 2024-09-11
feat: implement dedicated_server_credential data source

imtiazPabel:


internal/core/shared/enum_utils/enum.go
```diff
@@ -69,3 +69,22 @@ func ConvertStringEnumToValues[T fmt.Stringer](enumValues []T) []string {
 
 	return values
 }
+
+// ConvertStringSliceToValues expect any type of slice those are eventually string and return an slice of simple string.
+func ConvertStringSliceToValues[T ~string](enumValues []T) []string {
```
alrayyes, line 74: Nice. Can you check the existing enums to see where this code is duplicated so they all use the new convert functions?

internal/core/shared/enum_utils/enum_test.go
```diff
@@ -156,3 +156,24 @@ func ExampleConvertStringEnumToValues() {
 	fmt.Println(values)
 	// Output: [firstValue secondValue]
 }
+
+func TestConvertStringSliceToValues(t *testing.T) {
```
alrayyes, line 160: As these are exported functions it would be nice to have some example tests for the documentation. See any tests function that has an `Example` prefix.

internal/provider/dedicated_server_credential_data_source.go
```diff
@@ -0,0 +1,98 @@
...
+	Username          types.String `tfsdk:"username"`
+	Password          types.String `tfsdk:"password"`
+	Type              types.String `tfsdk:"type"`
+}
+
+func (d *dataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_dedicated_server_credential"
+}
+
+func (d *dataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
```
alrayyes, line 36: Can this be split up? Makes reviewing so much easier. Now I have to scan this whole file every time I want to compare code here to code in other files.

alrayyes (changes requested):
Please split up the files so I can review this properly on my teeny tiny laptop screen :p

internal/provider/dedicated_server_credential_data_source.go
```diff
@@ -0,0 +1,134 @@
...
+	_ datasource.DataSourceWithConfigure = &dataSource{}
+)
+
+type dataSource struct {
+	// TODO: Refactor this part, apiKey shouldn't be here.
+	apiKey string
+	client dedicatedServer.DedicatedServerAPI
+}
+
+type model struct {
```
majidkarimizadeh, line 27: since we are in the same package, we might need to be careful about these variables names, now it's fine but we can make it more specific.

majidkarimizadeh (approved):

arunvarghe (approved):

internal/provider/dedicated_server_credential_data_source.go
imtiazPabel: We are on a team agreement to do those things later.

internal/provider/dedicated_server_credential_data_source.go
imtiazPabel: You are right, I will change it on my next PR

PR #71, opened 2024-09-12
Revert leaseweb token back to leaseweb api token

mssiebelink:
Revert leaseweb token back to leaseweb api token
Improve Documentation


examples/provider/multiple_providers.tf
```diff
@@ -0,0 +1,28 @@
...
+    leaseweb = {
+      version = "0.3.3"
+      source  = "leaseweb/leaseweb"
+    }
+  }
+}
+
+provider "leaseweb" {
+  alias     = "nl"
+  api_token = "527070ca-8449-4f06-b609-ec6797bd8222"
```
imtiazPabel, line 12: Instead of version `0.3.3` we can specify some newer version as example, and it's now `token` instead of `api_token`

alrayyes (changes requested):
Make sure to enable linting by installing git hooks (`pnpm i`). It's a good idea to squash the commits into one that implements convention commit standard. Ie: `refactor!: rename api_token to leaseweb_api_token`. The `!` designates a breaking change.  See https://www.conventionalcommits.org/en/v1.0.0/ for the ful specs.

PR #73, opened 2024-09-17
feat: add dedicated server credential resource

majidkarimizadeh:


arunvarghe (approved):

internal/provider/dedicated_server_credential_resource.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (d *credentialResource) Schema(
+	_ context.Context,
+	_ resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"dedicated_server_id": schema.StringAttribute{
```
imtiazPabel, line 90: I believe for `server_id, username, type` should implement `RequiresReplace`. You can also check the legacy, at that time it was `ForceNew`

https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/force-new

internal/provider/dedicated_server_credential_resource.go
```diff
@@ -0,0 +1,247 @@
...
+		data.Password.ValueString(),
+		data.Type.ValueString(),
+		data.Username.ValueString(),
+	)
+	request := d.client.CreateServerCredential(d.authContext(ctx), data.DedicatedServerId.ValueString()).CreateServerCredentialOpts(*opts)
+	result, _, err := request.Execute()
+	if err != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf(
+				"Error creating credential for dedicated_server_id: %q",
```
imtiazPabel, line 131: Can you also mention the `username & server id`, as you did for other methods?

internal/provider/dedicated_server_credential_resource.go
majidkarimizadeh: Coooooool, in fact, I was looking for this! thanks

internal/provider/dedicated_server_credential_resource.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/leaseweb/leaseweb-go-sdk/dedicatedServer"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/provider/client"
+)
+
+var (
+	_ resource.Resource              = &credentialResource{}
+	_ resource.ResourceWithConfigure = &credentialResource{}
+)
+
+type credentialResource struct {
```
imtiazPabel, line 22: Not a deal breaker. 

`credentialResource` perhaps brings confusion, since we have also credential domain for `publicCloud`.
https://developer.leaseweb.com/api-docs/publiccloud_v1.html#tag/Credentials/operation/getCredentialList

Adding `dedicatedServer` in prefix, could be a solution, as we did for rest.

imtiazPabel (approved):

PR #78, opened 2024-09-23
feat: add dedicated server installation resource

imtiazPabel:


imtiazPabel:
@majidkarimizadeh, after picking the updated the `GO SDK`, there comes some breaking `ENUM` issues. I fixed and left a separate commit. Please let me know, if you wanna fix those in some different approach.

https://github.com/Leaseweb/terraform-provider-leaseweb/pull/78/commits/0011332c488791f52640fe944427c744e481abf1

internal/provider/dedicated_server_installation_resource.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+type dedicatedServerInstallationResource struct {
+	// TODO: Refactor this part, apiKey shouldn't be here.
+	apiKey string
+	client dedicatedServer.DedicatedServerAPI
+}
+
+type dedicatedServerInstallationResourceModel struct {
+	ID                types.String   `tfsdk:"id"`
```
majidkarimizadeh, line 43: can we decide on this rule to have Id instead of ID?
for other keys like `ControlPanelID` also let's have `ControlPanelId`!

https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/internal/provider/dedicated_server_resource.go#L29

imtiazPabel: Well, in my understanding, the other way around, like we should use `ControlPanelID` instead of `ControlPanelId`

Sharing some line from GO recommendation.
`This rule also applies to “ID”`
`so write “appID” instead of “appId”.`
`As an example: ServeHTTP not ServeHttp`
https://go.dev/wiki/CodeReviewComments

Please correct me, If i'm wrong.

internal/provider/dedicated_server_installation_resource.go
```diff
@@ -0,0 +1,464 @@
...
+	var SSHKeysList []string
+	for _, k := range plan.SSHKeys {
+		if valueStringOrNil(k) != nil {
+			SSHKeysList = append(SSHKeysList, k.ValueString())
+		}
+	}
+	SSHKeys := strings.Join(SSHKeysList, "\n")
+
+	opts := dedicatedServer.NewInstallOperatingSystemOpts(plan.OperatingSystemID.ValueString())
+	opts.CallbackUrl = valueStringOrNil(plan.CallbackURL)
```
majidkarimizadeh, line 354: How about
```golang
if !plan.CallbackURL.IsNull() && !plan.CallbackURL.IsUnknown() {
    opts.SetCallbackUrl(plan.CallbackURL)
}
```
?

I don't see any benefit to set everything nil and pass in request if they are not filled.

internal/provider/dedicated_server_installation_resource.go
```diff
@@ -0,0 +1,464 @@
...
+	opts.ControlPanelId = valueStringOrNil(plan.ControlPanelID)
+	opts.Device = valueStringOrNil(plan.Device)
+	opts.Hostname = valueStringOrNil(plan.Hostname)
+	opts.Partitions = partitions
+	opts.Password = valueStringOrNil(plan.Password)
+	opts.PostInstallScript = valueStringOrNil(plan.PostInstallScript)
+	opts.PowerCycle = valueBoolOrNil(plan.PowerCycle)
```
majidkarimizadeh, line 361:
```golang
if !plan.PowerCycle.IsNull() && !plan.PowerCycle.IsUnknown() {
    opts.SetPowerCycle(plan.PowerCycle)
}
```

IMH no need to be a separate function.

internal/provider/dedicated_server_installation_resource.go
```diff
@@ -0,0 +1,464 @@
...
+	}
+
+	serverID := plan.DedicatedServerID.ValueString()
+	result, response, err := r.client.InstallOperatingSystem(r.authContext(ctx), serverID).
+		InstallOperatingSystemOpts(*opts).Execute()
+
+	if err != nil {
+		resp.Diagnostics.AddError(
+			fmt.Sprintf("Error resource dedicated_server_installation for server id: %q", serverID),
+			extractErrorMessage(response, err),
```
majidkarimizadeh, line 375: Isn't better to for now to just print out all response if it's getting error?! 
I see you are also checking the error code 400 too and there is some custom logic and condition! 
IMH we can also see what other options do we have (by adding a TODO) and check for better/general solution about logging!

internal/provider/dedicated_server_installation_resource.go
majidkarimizadeh: same for `valueInt32OrNil`

internal/provider/dedicated_server_installation_resource.go
```diff
@@ -0,0 +1,464 @@
...
+		resp.Diagnostics.Append(diags...)
+		return
+	}
+	plan.Partitions = partitionsList
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
+
+}
+
+func (r *dedicatedServerInstallationResource) Read(_ context.Context, _ resource.ReadRequest, _ *resource.ReadResponse) {
```
majidkarimizadeh, line 417: Is it possible to use `panic("inimplemented") here for now if there is not any implementation for that?!`

internal/provider/dedicated_server_installation_resource.go
```diff
@@ -0,0 +1,464 @@
...
+
+	payload := result.GetPayload()
+	plan.ID = types.StringValue(result.GetUuid())
+	plan.Device = types.StringValue(payload.GetDevice())
+	plan.Timezone = types.StringValue(payload.GetTimezone())
+	plan.PowerCycle = types.BoolValue(payload.GetPowerCycle())
+
+	// Preparing partitions for the response
+	partitionsVal := []dedicatedServerInstallationPartitionsModel{}
+	for _, p := range payload.GetPartitions() {
```
majidkarimizadeh, line 388: isn't be better to merge these 2 for loops?
```golang
for _, p := range payload.GetPartitions() { ...
and 
for _, partition := range partitionsVal { ...
```

would be better for having more readable code IMO

internal/provider/dedicated_server_installation_resource.go
imtiazPabel: FYI, `error.Error()` only show you the `HTTP` code, when `HTTP` failed.

While trying to trigger the resource installation, I encountered an issue where, if an `HTTP` error occurs, `error.Error()` only shows the error code. As a `Terraform` developer, I couldn't fully understand what went wrong—imagine how much harder it would be for a customer. Keys like `device/hostname` and some other keys need to be correct in a way, but without the actual `HTTP` error message, it's difficult to pinpoint the issue.

That's why I implemented a check: if the error is significant, the HTTP error message is shown; otherwise, `error.Error()` suffices. I've marked all relevant functions with TODOs to move this logic elsewhere later.

I also believe that for the rest of our `data sources/resources`, we should display useful messages rather than just a code. However, we can address that later, we can use the method / approach different way.

internal/provider/dedicated_server_installation_resource.go
imtiazPabel: It's a good findings, indeed. I updated

internal/provider/dedicated_server_installation_resource.go
imtiazPabel: Although I couldn't find any events triggered beyond the `created` method.
I did attempt using `panic("unimplemented")`, which caused the test to fail. 
It doesn't seem particularly important at this point.

internal/provider/dedicated_server_installation_resource.go
imtiazPabel: It's not passing everything as `nil`, rather it's passing `value or nil(if no value)`. And that's what the `stuct` is expect, if you check.

In this file, I had to check whether the string has a value more than **15 times**. I'm not a fan of doing this repeatedly, especially when dealing with many similar fields.

IMHO, eliminating these repetitive null and unknown checks would make the code more declarative. Instead of multiple lines of logic for each field, we could have cleaner one-liner assignments.

internal/provider/dedicated_server_installation_resource.go
imtiazPabel: https://github.com/Leaseweb/terraform-provider-leaseweb/pull/78#discussion_r1771507156

internal/provider/dedicated_server_installation_resource.go
majidkarimizadeh: then sounds good, let's discuss it and implement it for the rest. later on and not in this PR :)

internal/provider/dedicated_server_installation_resource.go
majidkarimizadeh: True, value or nil. but why nil if it's not set or if it's empty string? it makes sense to not pass it at all through request

imtiazPabel: It makes sense to pass `value or nil`, because that’s the way `struct` builds, to receive data 😊

https://github.com/Leaseweb/leaseweb-go-sdk/blob/main/dedicatedServer/model_install_operating_system_opts.go#L22

arunvarghe: As Imtiaz mentioned, I see that struct accepts nil, So we can just pass nil also right? So It will reduce many if blocks with those additional methods.

arunvarghe (approved):

internal/provider/dedicated_server_installation_resource.go
alrayyes: I don't care. If you're both happy with ID i'll make it ID. But let me make another PR for that and change all Id references in publccloud.

PR #79, opened 2024-09-23
chore: change license from MPL 2.0 to Apache 2.0

imtiazPabel:


majidkarimizadeh:
based on https://www.hashicorp.com/license-faq
> The maintainers of each provider are free to select their own license.

Good to go

majidkarimizadeh (approved):

LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
majidkarimizadeh, line 190: @imtiazPabel we need to fill these

imtiazPabel: TBH, I also goes to fill these, but when checking what does others, I keep it as it is. 
I got confused, lmk if you've idea.

https://github.com/kubernetes/kubernetes/blob/master/LICENSE
https://github.com/tensorflow/tensorflow/blob/master/LICENSE
https://github.com/apache/kafka/blob/trunk/LICENSE

majidkarimizadeh (approved):

PR #81, opened 2024-09-26
[WIP] feat: add dedicated server resource

majidkarimizadeh:


internal/provider/dedicated_server_resource.go
```diff
@@ -26,35 +36,33 @@ type dedicatedServerResource struct {
...
-	CpuQuantity                        types.Int32  `tfsdk:"cpu_quantity"`
-	CpuType                            types.String `tfsdk:"cpu_type"`
+	ID                           types.String `tfsdk:"id"`
+	Reference                    types.String `tfsdk:"reference"`
+	ReverseLookup                types.String `tfsdk:"reverse_lookup"`
+	DHCPLease                    types.String `tfsdk:"dhcp_lease"`
+	PoweredOn                    types.Bool   `tfsdk:"powered_on"`
+	PublicNetworkInterfaceOpened types.Bool   `tfsdk:"public_network_interface_opened"`
+	PublicIpNullRouted           types.Bool   `tfsdk:"public_ip_null_routed"`
+	PublicIp                     types.String `tfsdk:"public_ip"`
```
imtiazPabel, line 46: Both,`Ip & Mac` is basically an acronym, which in that case would be `PublicIP & InternalMAC`

internal/provider/dedicated_server_resource.go
```diff
@@ -105,136 +113,116 @@ func (d *dedicatedServerResource) Configure(ctx context.Context, req resource.Co
...
-				Computed:    true,
-				Description: "The capacity of the rack.",
-			},
-			"rack_type": schema.StringAttribute{
+			"reference": schema.StringAttribute{
+				Optional:    true,
 				Computed:    true,
-				Description: "The type of the rack.",
+				Description: "Reference of server.",
+				PlanModifiers: []planmodifier.String{
```
imtiazPabel, line 130: I noticed that for every parameter, you've specified `UseStateForUnknown()`.

To share my understanding of `UseStateForUnknown()`, it's pretty self-explanatory. After the first `terraform apply`, when the resource is created, next subsequent` terraform plan/apply` it will pick values from the state.

The key difference I see is that using `UseStateForUnknown()` prevents displaying the `(known after apply)` message, which seems most appropriate for the `ID` field since an `ID` will typically remain unchanged once the resource is created.

However, other values could change depending on updates.

If my understanding is correct, you might want to consider using `UseStateForUnknown()` selectively, or please feel free to correct me if I'm mistaken.

internal/provider/dedicated_server_resource.go
```diff
@@ -243,266 +231,297 @@ func (d *dedicatedServerResource) Read(ctx context.Context, req resource.ReadReq
...
-	var locationRack, locationSite, locationSuite, locationUnit *string
-	if location, ok := result.GetLocationOk(); ok {
-		locationRack, _ = location.GetRackOk()
-		locationSite, _ = location.GetSiteOk()
-		locationSuite, _ = location.GetSuiteOk()
-		locationUnit, _ = location.GetUnitOk()
+	// Updating reference
+	if !plan.Reference.IsNull() && !plan.Reference.IsUnknown() {
+		ropts := dedicatedServer.NewUpdateServerReferenceOpts(plan.Reference.ValueString())
+		referenceRequest := d.client.UpdateServerReference(d.authContext(ctx), state.ID.ValueString()).UpdateServerReferenceOpts(*ropts)
```
imtiazPabel, line 281: Since we're not dealing with `referenceRequest`
can it be? `_, err := d.client.UpdateServerReference(d.authContext(ctx), state.ID.ValueString()).UpdateServerReferenceOpts(*ropts).Execute()`

internal/provider/dedicated_server_resource.go
```diff
@@ -243,266 +231,297 @@ func (d *dedicatedServerResource) Read(ctx context.Context, req resource.ReadReq
...
+				summary := fmt.Sprintf("Error null routing an IP for dedicated server: %q", state.ID.ValueString())
+				resp.Diagnostics.AddError(summary, err.Error())
+				return
+			}
+		} else {
+			nullRoutedRequest := d.client.RemoveNullIpRoute(d.authContext(ctx), state.ID.ValueString(), state.PublicIp.ValueString())
+			_, _, err := nullRoutedRequest.Execute()
+			if err != nil {
+				summary := fmt.Sprintf("Error remove null routing an IP for dedicated server: %q", state.ID.ValueString())
+				resp.Diagnostics.AddError(summary, err.Error())
```
imtiazPabel, line 342: Through schema, I tried to `removeNullIpRoute` and getting an error of `400 Bad Request`, couldn't understand what I did wrong. 

Then I did and `API` request manually, I found the message on error response `IP address not found`.
Isn't it important to show the useful message to customer?
On the update, it's calling a bunch of endpoints depends on the schema, otherwise I found it will be overwhelming for customer to figure it out, what wrong I did.

imtiazPabel:
A scenario, I run `terraform import leaseweb_dedicated_server.example 123456` on below schema

```
resource "leaseweb_dedicated_server" "example" {
  // nothing else
}
```

It's fine to import and create the state. Later I tried to update the reference like below

```
resource "leaseweb_dedicated_server" "example" {
  reference = "Update Name"
}
```

And it's went unsuccessful with the error `Error powering on for dedicated server: "123456"`. I know that, on `staging` environment, we can't test `power on/off`. But my concern is I didn't intend for any power changes then it shouldn't trigger this.

Comment out the block for power and found this happens for `IP null routing` as well without any intend of change.

internal/provider/dedicated_server_resource.go
```diff
@@ -243,266 +231,285 @@ func (d *dedicatedServerResource) Read(ctx context.Context, req resource.ReadReq
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading dedicated server power status with id: %q - %s", serverID, err.Error())
+	}
+	pdu := powerResult.GetPdu()
+	ipmi := powerResult.GetIpmi()
+	poweredOn := pdu.GetStatus() != "off" && ipmi.GetStatus() != "off"
+
+	// Getting server public network interface info
+	var publicNetworkOpened bool
+	networkRequest := d.client.GetNetworkInterface(d.authContext(ctx), serverID, dedicatedServer.NETWORKTYPE_PUBLIC)
```
imtiazPabel, line 469: `networkResult, networkResponse, err := d.client.GetNetworkInterface(d.authContext(ctx), serverID, dedicatedServer.NETWORKTYPE_PUBLIC).Execute()`

internal/provider/dedicated_server_resource.go
```diff
@@ -243,266 +231,285 @@ func (d *dedicatedServerResource) Read(ctx context.Context, req resource.ReadReq
...
-	var publicMac, publicIp, publicGateway *string
-	if networkInterfaces, ok := result.GetNetworkInterfacesOk(); ok {
-		if publicNetworkInterface, ok := networkInterfaces.GetPublicOk(); ok {
-			publicMac, _ = publicNetworkInterface.GetMacOk()
-			publicIp, _ = publicNetworkInterface.GetIpOk()
-			publicGateway, _ = publicNetworkInterface.GetGatewayOk()
+	// Updating Power status
+	if !plan.PoweredOn.IsNull() && !plan.PoweredOn.IsUnknown() {
+		if plan.PoweredOn.ValueBool() {
+			request := d.client.PowerServerOn(d.authContext(ctx), state.ID.ValueString())
```
imtiazPabel, line 293: https://github.com/Leaseweb/terraform-provider-leaseweb/pull/81#discussion_r1778282621
A couple of other similar place

internal/provider/dedicated_server_resource.go
majidkarimizadeh: That's a good one IMO, we can work on it but I'd suggest in a different ticket.

internal/provider/dedicated_server_resource.go
imtiazPabel: Not a deal breaker. But it still remains on a few places, would you like to change there also?

internal/provider/dedicated_server_resource.go
```diff
@@ -243,266 +212,285 @@ func (d *dedicatedServerResource) Read(ctx context.Context, req resource.ReadReq
...
 	}
 
-	var publicMac, publicIp, publicGateway *string
-	if networkInterfaces, ok := result.GetNetworkInterfacesOk(); ok {
-		if publicNetworkInterface, ok := networkInterfaces.GetPublicOk(); ok {
-			publicMac, _ = publicNetworkInterface.GetMacOk()
-			publicIp, _ = publicNetworkInterface.GetIpOk()
-			publicGateway, _ = publicNetworkInterface.GetGatewayOk()
+	// Updating Power status
+	if !plan.PoweredOn.IsNull() && !plan.PoweredOn.IsUnknown() {
```
imtiazPabel, line 272: Does, it seems, improve the readability and simplify by extracting common logic?

```
if !plan.PoweredOn.IsNull() && !plan.PoweredOn.IsUnknown() {
		var requestErr error
		serverID := state.ID.ValueString()

		if plan.PoweredOn.ValueBool() {
			_, requestErr = d.client.PowerServerOn(d.authContext(ctx), serverID).Execute()
		} else {
			_, requestErr = d.client.PowerServerOff(d.authContext(ctx), serverID).Execute()
		}

		if requestErr != nil {
			summary := fmt.Sprintf("Error powering %s for dedicated server: %q",
				map[bool]string{true: "on", false: "off"}[plan.PoweredOn.ValueBool()],
				serverID)
			resp.Diagnostics.AddError(summary, requestErr.Error())
			return
		}

		state.PoweredOn = plan.PoweredOn
	}
```

majidkarimizadeh: can we do this later?

imtiazPabel: If you found this helpful, there have some more places to refactor, like this way.

majidkarimizadeh: this is definitely helpfull! we'll look at all this stuff soon

imtiazPabel: @majidkarimizadeh, yes we can.

internal/provider/dedicated_server_resource.go
```diff
@@ -243,266 +212,285 @@ func (d *dedicatedServerResource) Read(ctx context.Context, req resource.ReadReq
...
-			publicMac, _ = publicNetworkInterface.GetMacOk()
-			publicIp, _ = publicNetworkInterface.GetIpOk()
-			publicGateway, _ = publicNetworkInterface.GetGatewayOk()
+	var remoteManagementIP string
+	if networkInterfaces, ok := serverResult.GetNetworkInterfacesOk(); ok {
+		if remoteNetworkInterface, ok := networkInterfaces.GetRemoteManagementOk(); ok {
+			remoteManagementIPPart := strings.Split(remoteNetworkInterface.GetIp(), "/")
+			ip := net.ParseIP(remoteManagementIPPart[0])
+			if ip != nil {
+				remoteManagementIP = ip.String()
```
imtiazPabel, line 422: For something later, I just shoot my comment, so that we can have a look, later.
Seems both `publicIP & remoteManagementIP` extractions are identical. If it is possible, we can extract the common logic.

imtiazPabel:
So far, I don’t see any more deal breakers from my side. There are a few points we can address later. However, I’m still a bit skeptical about this https://github.com/Leaseweb/terraform-provider-leaseweb/pull/81#discussion_r1778321534

Since this PR is tied to releasing a new tag, I’m unsure if it's a good decision to release without having useful error messages. After discussing within team, I think it would be fine to proceed.

imtiazPabel (approved):

arunvarghe (approved):

PR #82, opened 2024-10-02
chore: add error message helper function to show error message details

majidkarimizadeh:


internal/provider/error.go
```diff
@@ -0,0 +1,25 @@
+package provider
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+func getHttpErrorMessage(resp *http.Response, err error) string {
```
imtiazPabel, line 9: Basically, the method is not gonna response only HTTP error message. Depends on condition, it will respond `error.Error(ex: DEDICATED is not a valid RackType)`.

In that case doesn't make it misleading, the method name?

internal/provider/dedicated_server_credential_data_source.go
```diff
@@ -115,12 +115,12 @@ func (d *dedicatedServerCredentialDataSource) Read(ctx context.Context, req data
 	credType := dedicatedServer.CredentialType(data.Type.ValueString())
 	username := data.Username.ValueString()
 
-	credential, _, err := d.client.GetServerCredential(d.authContext(ctx), serverID, credType, username).Execute()
+	credential, response, err := d.client.GetServerCredential(d.authContext(ctx), serverID, credType, username).Execute()
 
 	if err != nil {
```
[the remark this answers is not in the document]
imtiazPabel: @alrayyes, both parts will be displayed in different locations when an error occurs.
**Heading:** `fmt.Sprintf("dedicated_server_credential %q", serverID)`
**Details:** `getHttpErrorMessage(response, err)`

Additionally, please note that `resp.Diagnostics.AddError` expects two parameters. Could you review your example code and refine it accordingly?

internal/provider/error.go
majidkarimizadeh: what do you suggest?! `getErrorMessage` is good?!

alrayyes: I would just call it 'NewError()' and return something that implements the golang error type. See https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/internal/repositories/shared/error.go & https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/internal/repositories/shared/error_response.go for an example.

imtiazPabel (approved):

alrayyes (approved):

PR #84, opened 2024-10-03
chore: poc from imtiaz

imtiazPabel:


imtiazPabel:
### The summary of the POC in points 
(Not a lot of changes 🤪)

1. **File Organization:**

     -    `getServer & update` two long methods are separated in 2 files, `get_server.go & update.go` respectively.
     -    I found schema and model are closely connected with together, that's why I kept them in a file `schema_model.go`.
     -    The remaining content is kept in `resource.go`

3. **Flexible Approach:**

     -   I prefer not to rigidly follow either always **splitting** files or always sticking to a **single** file. Instead, I believe the structure should depend on what enhances readability.
     
4. **Package Structure:**

      -   I plan to keep all dedicated server `resources / data sources` within a package, called `dedicatedserver`, for better organization.
     
5. **Removing Redundancy:**

     -   We’ve been repeating the `Configure` and `authContext` methods across all `data sources / resources` 🙄. To eliminate this redundancy, I created a struct called `API`, which I intend to use across the dedicatedserver `resources / data sources`

PR #103, opened 2024-10-15
Implement golang best practices for publiccloud package

alrayyes:
Move publiccloud logic to a more go friendly structure. Remove all the hexagonal design and go for a flat architecture.


internal/utils/pagination_test.go
```diff
@@ -1,13 +1,23 @@
-package shared
+package utils
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
 	"github.com/leaseweb/leaseweb-go-sdk/publicCloud"
```
majidkarimizadeh, line 8: there is no need of pagination to know about request, otherwise for dedicated server request and other APIs needs to be included here

alrayyes: This is the current implementation. I'd like to first focus on just the new structure before talking about refactoring functionality. This PR is big enough as it is. Right now only minimal changes have been made so everything works in the new structure.

alrayyes: I agree though, pagination should be agnostic. Indeed, we don't want to have to add stuff to the pagination module every time we want to use it with a new sdk.

alrayyes: Just check the pagination.go. Request.offset itself is updated, that's why it's passed.

internal/provider/publiccloud/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package publiccloud implements functionality for public_cloud data sources & resources.
```
imtiazPabel, line 1: Is this file automatically created?

alrayyes: No. this is used to update documentation on https://pkg.go.dev/github.com/leaseweb/terraform-provider-leaseweb and your ide when you hover over the package name.

internal/provider/publiccloud/instance_resource.go
```diff
@@ -0,0 +1,1398 @@
+package publiccloud
+
```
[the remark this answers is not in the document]
alrayyes: Isn't eh whole idea to keep logic that belongs together coupled in the same package?

internal/provider/publiccloud/instance_resource.go
```diff
@@ -0,0 +1,1398 @@
...
+	return r.Description(ctx)
+}
+
+func (r regionValidator) ValidateString(
+	_ context.Context,
+	request validator.StringRequest,
+	response *validator.StringResponse,
+) {
+	// If the region is unknown or null, there is nothing to validate.
+	if request.ConfigValue.IsUnknown() || request.ConfigValue.IsNull() {
```
imtiazPabel, line 144: This types of checking comes a lot of time over the terraform. Better to have a common/shared place.

imtiazPabel (approved):

imtiazPabel:
The PR is mainly about implementing the new structure. There are a few points, that I believe we can address later, but for now, it looks good to proceed the new directory structure.

PR #107, opened 2024-10-18
feat: add public cloud credential resource

majidkarimizadeh:
#### NOTE:
we need to decide about `utils.Error` function, for now I just moved it around to be able to use it for dedicated server and public cloud.


internal/provider/provider_test.go
```diff
@@ -855,3 +855,103 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+		instance_id = "695ddd91-051f-4dd6-9120-938a927a47d0"
+	   	username = "root"
+	   	type = "OPERATING_SYSTEM"
+	   	password = "12341234"
+	}`,
+					Check: resource.ComposeAggregateTestCheckFunc(
+						resource.TestCheckResourceAttr(
+							"leaseweb_public_cloud_credential.test",
+							"instance_id",
+							"695ddd91-051f-4dd6-9120-938a927a47d0",
```
alrayyes, line 877: What happens if a user inputs an invalid instance id?

internal/provider/provider_test.go
```diff
@@ -855,3 +855,103 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+			Steps: []resource.TestStep{
+				// Create and Read testing
+				{
+					Config: providerConfig + `
+	resource "leaseweb_public_cloud_credential" "test" {
+		instance_id = "695ddd91-051f-4dd6-9120-938a927a47d0"
+	   	username = "root"
+	   	type = "OPERATING_SYSTEM"
+	   	password = "12341234"
```
alrayyes, line 871: Just double checking, are there validation rules for passwords server side?

internal/provider/provider_test.go
```diff
@@ -855,3 +855,103 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+	t.Run("creates and updates a credential", func(t *testing.T) {
+		resource.Test(t, resource.TestCase{
+			ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+			Steps: []resource.TestStep{
+				// Create and Read testing
+				{
+					Config: providerConfig + `
+	resource "leaseweb_public_cloud_credential" "test" {
+		instance_id = "695ddd91-051f-4dd6-9120-938a927a47d0"
+	   	username = "root"
```
alrayyes, line 869: What happens when a user inputs a non existent user?

internal/provider/provider_test.go
```diff
@@ -855,3 +855,103 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+							"OPERATING_SYSTEM",
+						),
+						resource.TestCheckResourceAttr(
+							"leaseweb_public_cloud_credential.test",
+							"password",
+							"12341234",
+						),
+					),
+				},
+				// Update and Read testing
```
alrayyes, line 896: I'm guessing that because this is a credential importing is not possible?

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+var (
+	_ resource.Resource              = &credentialResource{}
+	_ resource.ResourceWithConfigure = &credentialResource{}
+)
+
+type credentialResource struct {
+	client client.Client
+}
+
+type credentialResourceData struct {
```
alrayyes, line 30: We need to come to an agreement as to what to name the model structs. Personally I like `resourceModelResource` as the prefix makes it clear what this is supposed to be. I think we should discuss this in a call so we're all on the same page.

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+func (c *credentialResource) Schema(
+	_ context.Context,
+	_ resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"instance_id": schema.StringAttribute{
+				Required:    true,
+				Description: `The ID of the instance.`,
```
alrayyes, line 80: The validation rules should be added to the description. I would to publiccloud to understand what the rules for these are. I assume for example that you can't create credentials for an instance that's stopped?

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+			"username": schema.StringAttribute{
+				Required:    true,
+				Description: `The username for the credentials`,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"type": schema.StringAttribute{
+				Required:    true,
+				Description: `The type of the credential. Valid options are: "OPERATING_SYSTEM", "CONTROL_PANEL"`,
```
alrayyes, line 94: Is there a publicCloud.Credential enum you can use here? Then you can be 100% sure that the documentation is correct in the future.

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+				Validators: []validator.String{
+					stringvalidator.OneOf([]string{"OPERATING_SYSTEM", "CONTROL_PANEL"}...),
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"password": schema.StringAttribute{
+				Required:    true,
+				Description: `The password for the credentials`,
```
alrayyes, line 104: I forgot what it's called, but I believe there is an attribute for secret data like this so the password isn't echoed to the user console or logged.

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (c *credentialResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
+	var data credentialResourceData
+	diags := req.Plan.Get(ctx, &data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	opts := publicCloud.NewStoreCredentialOpts(
```
alrayyes, line 118: It would be better to have this logic in a resource method. Easy to understand & test.

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+		publicCloud.CredentialType(data.Type.ValueString()),
+		data.Username.ValueString(),
+		data.Password.ValueString(),
+	)
+	request := c.client.PublicCloudAPI.StoreCredential(ctx, data.InstanceID.ValueString()).StoreCredentialOpts(*opts)
+	result, response, err := request.Execute()
+	if err != nil {
+		summary := fmt.Sprintf("Error creating credential with username: %q and instance_id: %q", data.Username.ValueString(), data.InstanceID.ValueString())
```
alrayyes, line 126: At some point we need a style guide for error messages.

internal/provider/publiccloud/resource_credential.go
```diff
@@ -0,0 +1,224 @@
...
+		resp.Diagnostics.AddError(summary, utils.NewError(response, err).Error())
+		tflog.Error(ctx, fmt.Sprintf("%s %s", summary, utils.NewError(response, err).Error()))
```
alrayyes, line 128: You can use the logging function in utils for this. Handles a lot of this for you.

alrayyes (changes requested):

internal/provider/publiccloud/resource_credential.go
imtiazPabel: IMHO, nice to be `credentialResourceModel` as terraform documentation have example. 
https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-resource-create#implement-resource-data-models

But, of course, could be anything else, if team wants to decide.

internal/provider/publiccloud/resource_credential.go
imtiazPabel: You are right. Nice to have logging in the same abstraction. Since we're gonna work later for error logging, at that time, maybe a lot of things gonna change, it's also fine to keep it like that.

internal/provider/publiccloud/resource_credential.go
majidkarimizadeh: added.

internal/provider/publiccloud/resource_credential.go
alrayyes: That's fine. But then at least add a TODO

internal/provider/provider_test.go
majidkarimizadeh: they are, although I also added rule to check the length

internal/provider/provider_test.go
majidkarimizadeh: server side validation will take care of that

internal/provider/provider_test.go
majidkarimizadeh: AFAIK no!

internal/provider/publiccloud/resource_credential.go
majidkarimizadeh: for logging, we need to discuss more, we need to show details of error message though!

majidkarimizadeh: btw TODO added.

internal/provider/publiccloud/resource_credential.go
majidkarimizadeh: It will be handled by server side validation

internal/provider/publiccloud/resource_credential.go
majidkarimizadeh: it's ONLY been used in here, no need to separate and move the login IMO!

internal/provider/provider_test.go
```diff
@@ -855,3 +855,127 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+	
+		resource "leaseweb_public_cloud_credential" "test" {
+			instance_id = "695ddd91-051f-4dd6-9120-938a927a47d0"
+		   	username = "root"
+		   	type = "invalid"
+		   	password = "12341234"
+		}`,
+
+						ExpectError: regexp.MustCompile(
+							`Attribute type value must be one of: \["OPERATING_SYSTEM" "CONTROL_PANEL"\],(\s*)got: "invalid"`,
```
alrayyes, line 974: Here it's better to just check for `Attribute type value must be one of:`. This way you don't need to refactor this test every time a type is added.

internal/provider/publiccloud/credential_resource.go
```diff
@@ -0,0 +1,228 @@
...
+	resp.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"instance_id": schema.StringAttribute{
+				Required:    true,
+				Description: `The ID of the instance.`,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"username": schema.StringAttribute{
```
alrayyes, line 85: If you want to check for an empty string here as well, but it's not a dealbreaker.

internal/provider/publiccloud/credential_resource.go
```diff
@@ -0,0 +1,228 @@
...
+				Required:    true,
+				Description: `The username for the credentials`,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"type": schema.StringAttribute{
+				Required:    true,
+				Description: `The type of the credential. Valid options are: "OPERATING_SYSTEM", "CONTROL_PANEL"`,
```
alrayyes, line 94: You can keep the documentation up to date by copying & using the helper in https://github.com/Leaseweb/terraform-provider-leaseweb/blob/1fd8e45c1f1db1714a81f30d88f2d104b259fa92/internal/provider/publiccloud/credential_data_source.go#L72.

internal/provider/publiccloud/credential_resource.go
```diff
@@ -0,0 +1,228 @@
...
+
+	data = credentialResourceModel{
+		InstanceID: data.InstanceID,
+		Type:       types.StringValue(string(result.GetType())),
+		Password:   types.StringValue(result.GetPassword()),
+		Username:   types.StringValue(result.GetUsername()),
+	}
+	diags = resp.State.Set(ctx, data)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
```
alrayyes, line 207: This can be removed as you're already at the end of the function

internal/provider/publiccloud/credential_resource.go
```diff
@@ -0,0 +1,228 @@
...
```
alrayyes, line 174: This can be removed.

alrayyes (changes requested):

alrayyes (approved):

arunvarghe (approved):

imtiazPabel (approved):

PR #108, opened 2024-10-21
Implement publiccloud image DataSource

alrayyes:


internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/leaseweb/leaseweb-go-sdk/publicCloud"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/provider/client"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/utils"
+)
+
+var (
+	_ datasource.DataSourceWithConfigure = &ImagesDataSource{}
```
imtiazPabel, line 18: I believe you forgot to add the interface `datasource.DataSource`
I couldn't find the struct as exported, if i'm right, then it should be `imagesDataSource`

alrayyes: On purpose. `DataSourceWithConfigure` already includes `DataSource` interface.

internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+type dataSourceModelImage struct {
+	Id           types.String `tfsdk:"id"`
```
imtiazPabel, line 22: My understanding says, `ID` instead of `Id` is the go recommendation. There was a discussion about it, here you can get it https://github.com/Leaseweb/terraform-provider-leaseweb/pull/78#discussion_r1771109552

alrayyes: I don't care. If we're going with ID I'll make it ID. But let's do that in another PR. I'll update all ID references in publiccloud.

internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+	Name         types.String `tfsdk:"name"`
+	Custom       types.Bool   `tfsdk:"custom"`
+	State        types.String `tfsdk:"state"`
+	MarketApps   []string     `tfsdk:"market_apps"`
+	StorageTypes []string     `tfsdk:"storage_types"`
+	Flavour      types.String `tfsdk:"flavour"`
+	Region       types.String `tfsdk:"region"`
+}
+
+func newDataSourceModelImageFromImage(sdkImage publicCloud.Image) dataSourceModelImage {
```
imtiazPabel, line 32: The name of the method more make sense to me `mapSDKImageToDataSourceImage / buildDataSourceImageFromSDKImage` or could be anything which is more readable. But `new` prefix doesn't sounds fit to me.

internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+	return dataSourceModelImage{
+		Id:      basetypes.NewStringValue(sdkImage.Id),
```
imtiazPabel, line 34: What if, we use the abstraction `sdkImage.GetId()`
Abstraction is doing some checks for us. Does it sounds helpful to you?

internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+		Name:    basetypes.NewStringValue(sdkImage.Name),
+		Custom:  basetypes.NewBoolValue(sdkImage.Custom),
+		Flavour: basetypes.NewStringValue(sdkImage.Flavour),
+	}
+}
+
+func newDataSourceModelImageFromImageDetails(
```
imtiazPabel, line 41: Same opinion for this and rest of the other methods from me.
https://github.com/Leaseweb/terraform-provider-leaseweb/pull/108/files#r1810408150

internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+	sdkImageDetails publicCloud.ImageDetails,
+) dataSourceModelImage {
+	return dataSourceModelImage{
+		Id:           basetypes.NewStringValue(sdkImageDetails.Id),
+		Name:         basetypes.NewStringValue(sdkImageDetails.Name),
+		Custom:       basetypes.NewBoolValue(sdkImageDetails.Custom),
+		State:        utils.AdaptNullableStringToStringValue(sdkImageDetails.State.Get()),
```
imtiazPabel, line 48: I think, that's how, you don't need to call a differnt method
basetypes.NewStringValue(sdkImageDetails.State.GetState())

internal/provider/publiccloud/images_datasource_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/leaseweb/leaseweb-go-sdk/publicCloud"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_newDataSourceModelImageFromImage(t *testing.T) {
```
imtiazPabel, line 13: IMHO, I couldn't find the test brings any value. It's all about you're repeating the implementation also on test and checking my repeating is right or not.

internal/provider/publiccloud/images_datasource.go
```diff
@@ -0,0 +1,233 @@
...
+		MarketApps:   sdkImageDetails.MarketApps,
+		StorageTypes: sdkImageDetails.StorageTypes,
+		Flavour:      basetypes.NewStringValue(sdkImageDetails.Flavour),
+		Region:       utils.AdaptNullableStringEnumToStringValue(sdkImageDetails.Region.Get()),
```
imtiazPabel, line 52: If it's enough to use `basetypes.NewStringValue(sdkImageDetails.GetRegion())`, then we can also remove the utils method

internal/provider/publiccloud/images_datasource.go
imtiazPabel: You are completely right. IDK, why I didn't check it and always use both of the interface 🤪

majidkarimizadeh:
pls reflect the architecture, then LGTM!

majidkarimizadeh (changes requested):
pls reflect the architecture, then LGTM!

majidkarimizadeh (approved):

arunvarghe (approved):

internal/provider/publiccloud/images_data_source.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/leaseweb/leaseweb-go-sdk/publicCloud"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/provider/client"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/utils"
+)
+
+var (
+	_ datasource.DataSourceWithConfigure = &imagesDataSource{}
+)
+
+type dataSourceModelImage struct {
```
imtiazPabel, line 21: Wouldn't it be `imageDataSourceModel`?

https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/ARCHITECTURE.md?plain=1#L26

imtiazPabel (approved):

PR #111, opened 2024-10-23
code quality changes

alrayyes:
- Rename Id to ID
- Use SDK model getters
- Rename model contractor functions


internal/provider/publiccloud/instance_resource.go
```diff
@@ -251,16 +251,16 @@ func (c resourceModelContract) IsContractTermValid() (bool, reason) {
...
-func newResourceModelContract(
+func adaptSdkContractToResourceContract(
 	_ context.Context,
 	sdkContract publicCloud.Contract,
 ) (*resourceModelContract, error) {
 	return &resourceModelContract{
-		BillingFrequency: basetypes.NewInt64Value(int64(sdkContract.BillingFrequency)),
-		Term:             basetypes.NewInt64Value(int64(sdkContract.Term)),
-		Type:             basetypes.NewStringValue(string(sdkContract.Type)),
+		BillingFrequency: basetypes.NewInt64Value(int64(sdkContract.GetBillingFrequency())),
```
majidkarimizadeh, line 259: no need to cast anymore!

alrayyes: There is. BillingFrequency is int32

internal/provider/publiccloud/instance_resource.go
```diff
@@ -491,12 +491,12 @@ func (i resourceModelInstance) CanBeTerminated(ctx context.Context) *reasonInsta
...
 
-func newResourceModelInstanceFromInstance(
+func adaptSdkInstanceToResourceInstance(
 	sdkInstance publicCloud.Instance,
 	ctx context.Context,
 ) (*resourceModelInstance, error) {
 	instance := resourceModelInstance{
-		Id:                  basetypes.NewStringValue(sdkInstance.Id),
+		ID:                  basetypes.NewStringValue(sdkInstance.Id),
 		Region:              basetypes.NewStringValue(string(sdkInstance.Region)),
```
majidkarimizadeh, line 500: `Reference:           basetypes.NewStringPointerValue(sdkInstance.Reference.Get())`????

internal/provider/publiccloud/instance_resource.go
```diff
@@ -491,12 +491,12 @@ func (i resourceModelInstance) CanBeTerminated(ctx context.Context) *reasonInsta
...
 		Reference:           utils.AdaptNullableStringToStringValue(sdkInstance.Reference.Get()),
```
majidkarimizadeh, line 501: `Reference:           basetypes.NewStringPointerValue(sdkInstance.Reference.Get())`

alrayyes: Good catch!

majidkarimizadeh (approved):

arunvarghe (approved):

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+
+type dataSourceModelContract struct {
+	BillingFrequency types.Int64  `tfsdk:"billing_frequency"`
+	Term             types.Int64  `tfsdk:"term"`
+	Type             types.String `tfsdk:"type"`
+	EndsAt           types.String `tfsdk:"ends_at"`
+	State            types.String `tfsdk:"state"`
+}
+
+func newDataSourceModelContract(sdkContract publicCloud.Contract) dataSourceModelContract {
```
imtiazPabel, line 23: Would you like to update also this method name accordingly?

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	return dataSourceModelContract{
+		BillingFrequency: basetypes.NewInt64Value(int64(sdkContract.GetBillingFrequency())),
+		Term:             basetypes.NewInt64Value(int64(sdkContract.GetTerm())),
+		Type:             basetypes.NewStringValue(string(sdkContract.GetType())),
+		EndsAt:           utils.AdaptNullableTimeToStringValue(sdkContract.EndsAt.Get()),
```
imtiazPabel, line 28: Wouldn't it be `sdkContract.GetEndsAt()` ?

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	Region              types.String            `tfsdk:"region"`
+	Reference           types.String            `tfsdk:"reference"`
+	Image               dataSourceModelImage    `tfsdk:"image"`
+	State               types.String            `tfsdk:"state"`
+	Type                types.String            `tfsdk:"type"`
+	RootDiskSize        types.Int64             `tfsdk:"root_disk_size"`
+	RootDiskStorageType types.String            `tfsdk:"root_disk_storage_type"`
+	Ips                 []dataSourceModelIp     `tfsdk:"ips"`
+	Contract            dataSourceModelContract `tfsdk:"contract"`
+	MarketAppId         types.String            `tfsdk:"market_app_id"`
```
imtiazPabel, line 44: Would it be `MarketAppID` and also `IPs`?

internal/provider/publiccloud/models.go
alrayyes: That does not work. That always returns the default value if nothing is set (a timestamp).

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+func adaptSdkInstanceToDatasourceInstance(sdkInstance publicCloud.Instance) dataSourceModelInstance {
+	var ips []dataSourceModelIp
+	for _, ip := range sdkInstance.Ips {
+		ips = append(ips, adaptSdkIpToDatasourceIp(ip))
+	}
+
+	return dataSourceModelInstance{
+		ID:                  basetypes.NewStringValue(sdkInstance.GetId()),
+		Region:              basetypes.NewStringValue(string(sdkInstance.GetRegion())),
+		Reference:           basetypes.NewStringPointerValue(sdkInstance.Reference.Get()),
```
imtiazPabel, line 56: `basetypes.NewStringValue(sdkInstance.GetReference())`

internal/provider/publiccloud/models.go
alrayyes: could do that

internal/provider/publiccloud/models.go
alrayyes: also does not work because GetReference() does not return nil, only an empty string.

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+		Contract:            newDataSourceModelContract(sdkInstance.GetContract()),
+		MarketAppId:         basetypes.NewStringPointerValue(sdkInstance.MarketAppId.Get()),
+	}
+}
+
+type dataSourceModelImage struct {
+	Id types.String `tfsdk:"id"`
+}
+
+func adaptSdkImageToDatasourceImage(sdkImage publicCloud.Image) dataSourceModelImage {
```
imtiazPabel, line 72: It's just a one-liner thing.
First question is it really needed a method?
If needed then only passing the imageID should be enough. No usage for the `publicCloud.Image` struct

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	return dataSourceModelImage{
+		Id: basetypes.NewStringValue(sdkImage.Id),
+	}
+}
+
+type dataSourceModelIp struct {
+	Ip types.String `tfsdk:"ip"`
+}
+
+func adaptSdkIpToDatasourceIp(sdkIp publicCloud.Ip) dataSourceModelIp {
```
imtiazPabel, line 82: Same opinion https://github.com/Leaseweb/terraform-provider-leaseweb/pull/111/files#r1812432649

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	return dataSourceModelIp{
+		Ip: basetypes.NewStringValue(sdkIp.Ip),
```
imtiazPabel, line 84: As you're refactoring with `get()` abstraction. Nice to use `sdkIp.getIP()` instead. Have some more place, like this way.

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	_ context.Context,
+	sdkContract publicCloud.Contract,
+) (*resourceModelContract, error) {
+	return &resourceModelContract{
+		BillingFrequency: basetypes.NewInt64Value(int64(sdkContract.GetBillingFrequency())),
+		Term:             basetypes.NewInt64Value(int64(sdkContract.GetTerm())),
+		Type:             basetypes.NewStringValue(string(sdkContract.GetType())),
+		EndsAt:           utils.AdaptNullableTimeToStringValue(sdkContract.EndsAt.Get()),
+		State:            basetypes.NewStringValue(string(sdkContract.GetState())),
+	}, nil
```
imtiazPabel, line 151: Curious minds, wants to know why have to return `nil`?

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+type resourceModelImage struct {
+	Id types.String `tfsdk:"id"`
```
imtiazPabel, line 155: `ID`

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+}
+
+func (i resourceModelImage) AttributeTypes() map[string]attr.Type {
+	return map[string]attr.Type{
+		"id": types.StringType,
+	}
+}
+
+func adaptSdkImageToResourceImage(
+	_ context.Context,
```
imtiazPabel, line 165: What value brings by passing `context.Context`?
I can't see any use case.

internal/provider/publiccloud/models.go
alrayyes: It is needed. In the image resource PR more stuff is added.

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	sdkImage publicCloud.Image,
+) (*resourceModelImage, error) {
+	return &resourceModelImage{
+		Id: basetypes.NewStringValue(sdkImage.Id),
+	}, nil
+}
+
+type reasonInstanceCannotBeTerminated string
+
+type resourceModelInstance struct {
```
imtiazPabel, line 175: `dataSourceModelInstance` and `resourceModelInstance` both struct are identical. Is it possible to reuse?

internal/provider/publiccloud/models.go
alrayyes: In this case yes, this isn't needed....yet..... I'll remove the function as it indeed does nothing.

internal/provider/publiccloud/models.go
alrayyes: I missed that one.

internal/provider/publiccloud/models.go
alrayyes: It returns `basetypes.NewStringNull()` if value is nil.

internal/provider/publiccloud/models.go
alrayyes: I find this very counterintuitive. But if this is the go way, it's a the go way.

internal/provider/publiccloud/models.go
alrayyes: Here none,, that's correct. But when you want to convert models to object types the terraform functions expect context to be passed, so it's needed then.

So the adapter function in value_adapters also passed context.

internal/provider/publiccloud/models.go
alrayyes: those change in future PRs.

But even if they don't, let's please not. That will make things very confusing very quickly. Even if two models are exactly the same they're 2 different things.

internal/provider/publiccloud/models.go
```diff
@@ -0,0 +1,508 @@
...
+	return opts, nil
+}
+
+func (i resourceModelInstance) GetUpdateInstanceOpts(ctx context.Context) (
+	*publicCloud.UpdateInstanceOpts,
+	error,
+) {
+	opts := publicCloud.NewUpdateInstanceOpts()
+
+	opts.Reference = i.Reference.ValueStringPointer()
```
imtiazPabel, line 294: `ValueStringPointer` checks only `isNull` and pass pointer. What if the value `IsUnknown`, then it will fail to initialize

alrayyes: Indeed. There was an adapter for that as well but I removed it in this PR. All the tests pass here. If there are issues in the future I will re add the adapter.

internal/provider/publiccloud/validators.go
```diff
@@ -0,0 +1,190 @@
...
+- state is equal to Creating
+- state is equal to Destroying
+- state is equal to Destroyed
+- contract.endsAt is set
+
+In all other scenarios an instance can be terminated.
+`
+}
+
+func (i instanceTerminationValidator) MarkdownDescription(ctx context.Context) string {
```
imtiazPabel, line 74: Can we use like reference `func (i * instanceTerminationValidator)`?

alrayyes: not needed here....yet.....

goland tells when this is needed. PS: I highly recommend installing goland.

alrayyes: Cannot do this

`cannot use instanceTerminationValidator{} (value of type instanceTerminationValidator) as validator.Object value in variable declaration: instanceTerminationValidator does not implement validator.Object (method Description has pointer receiver)

internal/provider/publiccloud/instance_resource.go
```diff
@@ -491,132 +300,132 @@ func (i resourceModelInstance) CanBeTerminated(ctx context.Context) *reasonInsta
...
 }
 
-func newResourceModelInstanceFromInstance(
+func adaptSdkInstanceToResourceInstance(
 	sdkInstance publicCloud.Instance,
 	ctx context.Context,
 ) (*resourceModelInstance, error) {
 	instance := resourceModelInstance{
-		Id:                  basetypes.NewStringValue(sdkInstance.Id),
+		ID:                  basetypes.NewStringValue(sdkInstance.Id),
```
imtiazPabel, line 308: Since the refactor is also about using SDK model getters. Can it be `sdkInstance.getId()`
A lot of more place have like this.

internal/provider/publiccloud/instance_resource.go
```diff
@@ -491,132 +300,132 @@ func (i resourceModelInstance) CanBeTerminated(ctx context.Context) *reasonInsta
...
+		return nil, fmt.Errorf("adaptSdkInstanceToResourceInstance: %w", err)
 	}
 	instance.Contract = contract
 
 	return &instance, nil
 }
 
-type resourceModelIp struct {
+type resourceModelIP struct {
 	Ip types.String `tfsdk:"ip"`
```
imtiazPabel, line 406: Can it also be `IP`?

internal/provider/publiccloud/instance_resource.go
```diff
@@ -491,132 +300,132 @@ func (i resourceModelInstance) CanBeTerminated(ctx context.Context) *reasonInsta
...
-func (i resourceModelIp) AttributeTypes() map[string]attr.Type {
+func (i resourceModelIP) AttributeTypes() map[string]attr.Type {
 	return map[string]attr.Type{
 		"ip": types.StringType,
 	}
 }
 
-func newResourceModelIpFromIp(
+func adaptSdkIpToResourceIP(
 	_ context.Context,
```
imtiazPabel, line 416: As we've discussed, `context` in param and `nil` is return aren't part of this method.

imtiazPabel (approved):

internal/provider/publiccloud/models.go
majidkarimizadeh: that's my question too! but we'll think about it later

majidkarimizadeh (approved):

PR #114, opened 2024-10-24
Fix camelCase in datasourceModel files

alrayyes:


majidkarimizadeh (changes requested):
pls reflect the architecture, then LGTM!

PR #115, opened 2024-10-24
feat: implement public_cloud_credential data source

imtiazPabel:
Similar to Majid's PR, we need to make a decision regarding the `utils.Error` function. For now, I've moved it to enable its use for the `public_cloud_credential` data source.


majidkarimizadeh:
LGTM, pls reflect the architecture (file naming) then we're good to go.

majidkarimizadeh (approved):

internal/provider/publiccloud/credential_data_source.go
```diff
@@ -0,0 +1,117 @@
...
+
+type credentialDataSource struct {
+	client publicCloud.PublicCloudAPI
+}
+
+func NewCredentialDataSource() datasource.DataSource {
+	return &credentialDataSource{}
+}
+
+type credentialDataSourceModel struct {
```
alrayyes, line 30: It makes more sense to move this above `credentialDataSource` , and keep the struct functions (Config, Read etc) close to the struct they're members of.

internal/utils/enum.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+// EnumToSlice converts any slice of custom types (enums) that are underlying string types to a slice of strings.
+func EnumToSlice[T ~string](enumValues []T) []string {
```
alrayyes, line 4: You just copied an existing adapter https://github.com/Leaseweb/terraform-provider-leaseweb/blob/1fd8e45c1f1db1714a81f30d88f2d104b259fa92/internal/utils/value_adapters.go#L127

internal/utils/enum.go
```diff
@@ -0,0 +1,19 @@
...
+	strValues := make([]string, len(enumValues))
+	for i, v := range enumValues {
+		strValues[i] = string(v)
+	}
+	return strValues
+}
+
+// EnumToMarkdown converts any slice of custom types (enums) that are underlying string types to markdown string.
+func EnumToMarkdown[T ~string](enumValues []T) string {
```
alrayyes, line 13: It makes sense to rename `int_markdown_list.go` to `markdown_helpers.go` and move this function there.

internal/utils/enum_test.go
```diff
@@ -0,0 +1,37 @@
...
+	want := []string{
+		"TEST_ONE",
+		"TEST_TWO",
+	}
+
+	got := EnumToSlice(enumValues)
+	assert.Equal(t, want, got)
+}
+
+func TestEnumToMarkdown(t *testing.T) {
```
alrayyes, line 26: Could you add an example test? Here you go:

```go
func ExampleEnumToMarkdown() {
	list := EnumToMarkdown(publicCloud.AllowedContractTypeEnumValues)

	fmt.Println(list)
	/**
	  Output:
	  - *HOURLY*
	  - *MONTHLY*
	*/
}

```

internal/provider/publiccloud/credential_data_source.go
```diff
@@ -0,0 +1,117 @@
...
+			},
+			"type": schema.StringAttribute{
+				Required:    true,
+				Description: "The type of the credential. Valid options are " + utils.EnumToMarkdown(publicCloud.AllowedCredentialTypeEnumValues),
+				Validators: []validator.String{
+					stringvalidator.OneOf(utils.EnumToSlice(publicCloud.AllowedCredentialTypeEnumValues)...),
+				},
+			},
+			"username": schema.StringAttribute{
+				Description: "The username for the credentials",
```
alrayyes, line 78: `Credentials username` sounds better. Same for password.

internal/provider/publiccloud/credential_data_source_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_credentialDataSource_Metadata(t *testing.T) {
```
alrayyes, line 11: This can be safely removed as the acceptance tests will all fail if metadata is incorrect.

internal/provider/provider_test.go
```diff
@@ -855,3 +855,129 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+							"The argument \"instance_id\" is required, but no definition was found",
+						),
+					},
+				},
+			})
+		},
+	)
+
+	t.Run(
+		"type should be in the request",
```
alrayyes, line 921: for tests language like `type is required` makes more sense.

internal/provider/provider_test.go
```diff
@@ -855,3 +855,129 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+								"12341234",
+							),
+						),
+					},
+				},
+			})
+		})
+
+	t.Run(
+		"instance id should be in the request",
```
alrayyes, line 900: instance_id

internal/provider/provider_test.go
```diff
@@ -855,3 +855,129 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+							"The argument \"username\" is required, but no definition was found",
+						),
+					},
+				},
+			})
+		},
+	)
+
+	t.Run(
+		"type should be one of these values OPERATING_SYSTEM CONTROL_PANEL",
```
alrayyes, line 963: As type can change over time, would be better to put something like `invalid type is not accepted`

internal/provider/provider_test.go
```diff
@@ -855,3 +855,129 @@ resource "leaseweb_public_cloud_instance" "test" {
...
+				Steps: []resource.TestStep{
+					{
+						Config: providerConfig + `
+        data "leaseweb_public_cloud_credential" "test" {
+          instance_id         = "695ddd91-051f-4dd6-9120-938a927a47d0"
+          type                = "A_WRONG_TYPE"
+          username            = "root"
+        }`,
+						ExpectError: regexp.MustCompile(
+							`Attribute type value must be one of: \["OPERATING_SYSTEM" "CONTROL_PANEL"\]`,
```
alrayyes, line 976: Same here. To avoid future test fixing it's a good idea to replace this with `Attribute type value must be one of:` and leave out the specifics.

alrayyes (changes requested):

arunvarghe (approved):

internal/provider/publiccloud/credential_data_source.go
imtiazPabel: I believe, there is no convention no community standards from go, neither terraform like that.

In fact, the terraform documentation organize this way. This is very trivial, and not something to discuss and come on agreement IMHO.

internal/utils/enum.go
imtiazPabel: I haven't noticed. Thanks for informing.

internal/provider/publiccloud/credential_data_source.go
alrayyes: I don't know if there is or isn't a convention for this. You can structure the file any way you like, but for sanity (not to mention clean merge diffs) it makes sense to group functions that belong to a struct underneath the struct. This also makes reviewing way easier.

alrayyes (approved):

PR #124, opened 2024-10-29
Map API validation to resource attributes

alrayyes:
- Implement `SetAttributeErrorsFromServerResponse` function to easily map api validation errors to attributes. If an attribute cannot be found, the error is shown on a resource level.
- Remove all custom validation from public_cloud to thoroughly test `SetAttributeErrorsFromServerResponse`.
- Move `LogError` function to `error.go`. 
- Remove `SdkError` struct as it's not needed anymore.

`SetAttributeErrorsFromServerResponse` can easily be implemented everywhere. All you need to do is pass the summary,  http response & terraform response diagnostics.. The function takes care of the rest.


internal/utils/error.go
```diff
@@ -13,11 +22,17 @@ type Error struct {
...
 // TODO: we need to merge error.go and log_error.go and have unified error/logging functionality.
 func (e Error) Error() string {
-	// Check if response or its body is nil, or if the status code is not an error.
+	// Check if the response or its body is nil,
+	//or if the status code is not an error.
 	if e.resp == nil || e.resp.Body == nil || e.resp.StatusCode < 400 {
 		return e.err.Error()
 	}
-	defer e.resp.Body.Close()
+	defer func(Body io.ReadCloser) {
```
alrayyes, line 30: If there was an error here the provider would fail. Now we know why at least.

majidkarimizadeh:
pls attach a screenshot of the error you show to customer, need to see how it looks from customer perspective.

alrayyes:
> pls attach a screenshot of the error you show to customer, need to see how it looks from customer perspective.

2024-10-30T09:01:20,980577407+01:00

First one is mapped. The second one is what's shown when mapping can't be done.

internal/utils/error.go
```diff
@@ -75,3 +90,118 @@ func NewError(resp *http.Response, err error) Error {
...
+		LogError(ctx, response, summary)
+		return
+	}
+
+	// Convert server response to ErrorResponse object.
+	errorResponse, err := newErrorResponse(response.Body)
+	// If error cannot be translated,
+	// Terraform will show a general error to the user.
+	if err != nil {
+		LogError(ctx, response, summary)
```
imtiazPabel, line 134: I found an observation about there and other places where we immediately `LogError and return`.

Think if condition is `true`, at that time it might write the error log but there is no `resp.Diagnostics` for the customer. As a result, customer gonna get issue with `Error: Missing Resource State After Create`

We should through `resp.Diagnostics`, if we want to stop there for customer.

For your testing help

keep those line anywhere in the file 

```
type testError struct{}

func (e *testError) Error() string {
	return "this is a test error"
}
```

and re-initialize the `err` value like this, to run the block inside if condition `err = &testError{}`

If it sounds complicated, what I mean, please let me know.

imtiazPabel:
For `image_resource`, I noticed there’s an `ID` parameter, but the `API` indicates we actually need an `instanceID` to create the image, as well as a separate `imageID` from the response.

If I'm correct, it seems we’re currently using the same `ID` for both `instanceID` and `imageID`, which might have been overlooked 🤪.

I also noticed that the updated error message doesn’t fully align with our last agreement. Since I’m assigned to this story, feel free to leave that part, and I’ll handle it after your merge.

Otherwise, it's a solid implementation—the custom validation is streamlined, and the `resource/data source` is much cleaner!

LGTM, kudos for your effort.

internal/utils/error.go
```diff
@@ -64,3 +81,115 @@ func NewError(resp *http.Response, err error) Error {
...
+	ErrorMessage  string              `json:"errorMessage"`
+	ErrorDetails  map[string][]string `json:"errorDetails"`
+}
+
+func handleError(
+	summary string,
+	err error,
+	diags *diag.Diagnostics,
+) {
+	const pleaseFileABugReport = "An error has occurred, please submit a bug report."
```
imtiazPabel, line 169: There is a const on this file `DefaultErrMsg` which is already using across the project.

internal/provider/publiccloud/image_resource.go
```diff
@@ -337,28 +299,22 @@ func (i *imageResource) Update(
...
 	).UpdateImageOpts(opts).Execute()
 	if err != nil {
-		sdkErr := utils.NewSdkError("", err, apiResponse)
 		summary := fmt.Sprintf(
 			"Updating resource %s for id %q",
 			i.name,
 			plan.ID.ValueString(),
 		)
-		response.Diagnostics.AddError(
+		utils.HandleSdkError(
```
imtiazPabel, line 312: This method is satisfying our agreed error Details format `utils.NewError(apiResponse, err).Error()`
https://jira.ocom.com/browse/DEVP-6193

Method `HandleSdkError` should also have some checks like, if it founds that the `httpError` is not related anything for `validation`, then should response regular error details.

Ex. this is an expected general http error, if it's not related with validation 

```
{
│   "correlationId": "865f5c22-5990-4313-a7a5-d2b8b302c94c",
│   "errorCode": "404",
│   "errorMessage": "Server not found"
│ }
```

In this case ATM, we are only getting `404 Not Found`

imtiazPabel:
LGTM I do have some plans to clean up the error log/messages and explore possible improvements, though I’m unsure how much we can achieve. Since we're unifying the error logging method, I think we can address those on the next PR. 

For this PR, the focus is primarily on validation stuff, and that part is working well.

imtiazPabel (approved):

PR #125, opened 2024-10-30
Implement workaround so dependabot updates docker images

alrayyes:
dependabot [does not support docker-compose](https://github.com/dependabot/dependabot-core/issues/390). As a workaround for now use Dockerfiles which will get automatically updated by dependabot.


docker-compose.yml
```diff
@@ -1,6 +1,7 @@
 services:
   golangci-lint:
-    image: golangci/golangci-lint:v1.58.2-alpine@sha256:511101feba931bd578164b77714d1cee015d8980e0d40033c378b6626193e4ab
+    build:
+      dockerfile: docker/golangci/Dockerfile
```
arunvarghe, line 4: Just to understand!! what was the reason to pull base image from Dockerfile?

arunvarghe: Okay, I understood, you want dependabot to update base image version.

alrayyes: indeed

arunvarghe (approved):

PR #137, opened 2024-11-06
fix: implement instance_id attribute to create image resource

alrayyes:
- instance_id is now used instead of id for image creation
- importing is now disabled as there is no way to know what the instance_id is for any imported image as that value is not returned by the api


docs/resources/public_cloud_instance.md
```diff
@@ -170,6 +170,7 @@ Read-Only:
 
 - `custom` (Boolean) Standard or Custom image
 - `flavour` (String)
+- `instance_id` (String) Can be either an Operating System or a UUID in case of a Custom Image ID.
```
imtiazPabel, line 173: Isn't it `image_id`?

alrayyes: no it's not.  It has to be instance_id because that's what's in create https://developer.leaseweb.com/api-docs/publiccloud_v1.html#tag/Images/operation/createImage. ImageId is  instance.Image.Id

imtiazPabel: I amn't so clear, while you're on `instance` resource, there already have a field `id` then why again `instance_id`?

Although, the description(`Can be either an Operating System or a UUID in case of a Custom Image ID`) I can only see for the `image_id` in api-defination.

alrayyes: because it uses the image resource model.

imtiazPabel: You're on `instance_resource`, they why the relation gonna come with `image_resource` and also on the file `instance_resource.go` only the key is added in schema, no other changes I can see.

alrayyes: I've removed instance_id from the loadBalancer resource schema. Unfortunately the image schema can't be reused in loadBalancer schema because there are other differences.

PR #138, opened 2024-11-06
fix: use returned request so new offset is properly set

alrayyes:


imtiazPabel (approved):

internal/provider/publiccloud/images_data_source.go
```diff
@@ -109,7 +109,7 @@ func getAllImages(
 			break
 		}
 
-		request.Offset(*offset)
+		request = request.Offset(*offset)
```
imtiazPabel, line 112: Instead of reassign, did you tried to update the reference?
IDK, is it possible or not.

PR #143, opened 2024-11-08
[WIP] Refactor/unified error method

imtiazPabel:
For the time beings, I've written the new function in the `new-error.go`, which I'm going to move `error.go` after get rid of all the things from `error.go`

Since there's already and `Error` method, I've named the new one `Errors` temporarily to avoid breaking the existing system.

I'm currently working on adding test cases, afterward, I'll apply the new method throughout the project.

ATM, I wanted to sync with everyone that this is the approach I'm taking. If anyone has suggestions, please let me know. 

I believe this will make PR smoother to close and help minimize any back & forth.


imtiazPabel:
I noticed that sometimes we miss the proper format when writing the error summary, so I wrote a method to mitigate the confusion.

The entire changes for the new method are solely on the commit, so that, if we're not agreed, it's easily can get rid of https://github.com/Leaseweb/terraform-provider-leaseweb/pull/143/commits/b4ce89efc7887d977bcda862056a6968eb8a3ecf

majidkarimizadeh:
> I noticed that sometimes we miss the proper format when writing the error summary, so I wrote a method to mitigate the confusion.
> 
> The entire changes for the new method are solely on the commit, so that, if we're not agreed, it's easily can get rid of [b4ce89e](https://github.com/Leaseweb/terraform-provider-leaseweb/commit/b4ce89efc7887d977bcda862056a6968eb8a3ecf)

IMO we don't need that function because at least to me, it's not clear what to pass every time as arguments! using `sprintf` would be enough

internal/utils/error.go
```diff
@@ -17,194 +16,106 @@ import (
...
-// that HandleSdkError can handle.
-// `instanceId` & `instance.id` both become `instance_id`.
-func normalizeErrorResponseKey(key string) string {
-	// Assume that the key has the format `contract.id`
-	//if any dots are found.
-	if strings.Contains(key, ".") {
-		return strings.ToLower(strings.Replace(key, ".", "_", -1))
+	if e.diags == nil {
+		e.writeLog("unable to record error details.")
+		log.Fatal(e.summary, DefaultErrMsg)
```
alrayyes, line 58: This is the way.

alrayyes (approved):

internal/provider/dedicatedserver/contract_panels_data_source.go
```diff
@@ -96,9 +95,8 @@ func (c *controlPanelsDataSource) Read(
 	}
 
 	if err != nil {
-		summary := fmt.Sprintf("Reading data %s", c.name)
-		resp.Diagnostics.AddError(summary, utils.NewError(response, err).Error())
-		tflog.Error(ctx, fmt.Sprintf("%s %s", summary, utils.NewError(response, err).Error()))
+		summary := utils.BuildSummary(c.name, "Reading data")
```
alrayyes, line 98: It would be more clear to the developer if the second argument was an enum. That way BuildSummary can be used for standard stuff (read/update/delete) which covers 90% of use cases and the developer can use sfprintf for the other 10%

alrayyes:
> > I noticed that sometimes we miss the proper format when writing the error summary, so I wrote a method to mitigate the confusion.
> > The entire changes for the new method are solely on the commit, so that, if we're not agreed, it's easily can get rid of [b4ce89e](https://github.com/Leaseweb/terraform-provider-leaseweb/commit/b4ce89efc7887d977bcda862056a6968eb8a3ecf)
> 
> IMO we don't need that function because at least to me, it's not clear what to pass every time as arguments! using `sprintf` would be enough

See https://github.com/Leaseweb/terraform-provider-leaseweb/pull/143/files#r1836126193. We can do both.

alrayyes:
@imtiazPabel Did you check the code coverage to make sure the tests don't miss anything?

imtiazPabel:
> > > I noticed that sometimes we miss the proper format when writing the error summary, so I wrote a method to mitigate the confusion.
> > > The entire changes for the new method are solely on the commit, so that, if we're not agreed, it's easily can get rid of [b4ce89e](https://github.com/Leaseweb/terraform-provider-leaseweb/commit/b4ce89efc7887d977bcda862056a6968eb8a3ecf)
> > 
> > 
> > IMO we don't need that function because at least to me, it's not clear what to pass every time as arguments! using `sprintf` would be enough
> 
> See https://github.com/Leaseweb/terraform-provider-leaseweb/pull/143/files#r1836126193. We can do both.

Since `buildSummary` is an extra feature and not the primary focus of this PR, I removed the commit to keep things straightforward.

However, since I've already put in some work, I've passed the [commit to a different branch](https://github.com/imtiazPabel/terraform-provider-leaseweb/compare/refactor/unified-error-method...imtiazPabel:terraform-provider-leaseweb:feat/method-build-summary?expand=1). This way, if we decide later that something like this is needed, we can pick up from there.

imtiazPabel:
> @imtiazPabel Did you check the code coverage to make sure the tests don't miss anything?

In my understanding yes. However, if you've got any concerns, please let me know.

alrayyes:
> > @imtiazPabel Did you check the code coverage to make sure the tests don't miss anything?
> 
> In my understanding yes. However, if you've got any concerns, please let me know.

Just checking. This code is used everywhere and has a bunch of edge cases. Therefore it's important that it's tested properly. If you copied my tests it should be OK.

imtiazPabel:
> > > @imtiazPabel Did you check the code coverage to make sure the tests don't miss anything?
> > 
> > 
> > In my understanding yes. However, if you've got any concerns, please let me know.
> 
> Just checking. This code is used everywhere and has a bunch of edge cases. Therefore it's important that it's tested properly. If you copied my tests it should be OK.

I didn’t copy your test 😁 instead, I aimed to cover every case in the tests and also did some manual testing across all scenarios.

PR #147, opened 2024-11-08
POC add helper to generate meta type name

alrayyes:
A couple of utility functions to generate metadata type names. Advantages:

- One single place where we format all the names
- Avoid user error. It's very easy to make a mistake that isn't caught by a human here
- One less thing to worry about during code reviews


imtiazPabel:
In my opinion, we could completely get rid of the `metadata` and `configure` method from all the `data source / resource`. Both method are identical.

In one of a PR, I showed, how we can achieve it, just creating a base or common struct. https://github.com/Leaseweb/terraform-provider-leaseweb/pull/84/files#diff-82a3e70b00afd9270c2d5a4ed5953146891d727f4cbe2d91031b8e63fe7f75adR12

https://github.com/Leaseweb/terraform-provider-leaseweb/pull/84#issuecomment-2391545303 (Check Removing Redundancy part)

Which we can grow more, if there comes more redundancy like `generateSummary` is also something like that.
Subject to team agreement

alrayyes:
> In my opinion, we could completely get rid of the `metadata` and `configure` method from all the `data source / resource`. Both method are identical.
> 
> In one of a PR, I showed, how we can achieve it, just creating a base or common struct. https://github.com/Leaseweb/terraform-provider-leaseweb/pull/84/files#diff-82a3e70b00afd9270c2d5a4ed5953146891d727f4cbe2d91031b8e63fe7f75adR12
> 
> [#84 (comment)](https://github.com/Leaseweb/terraform-provider-leaseweb/pull/84#issuecomment-2391545303) (Check Removing Redundancy part)
> 
> Which we can grow more, if there comes more redundancy like `generateSummary` is also something like that. Subject to team agreement

That is indeed better and should work without issues.

PR #148, opened 2024-11-08
feat: add helper to get client

alrayyes:
A couple of utility functions to configure the client. Advantages:

- One single piece of logic to setup
- Avoid user error. It's very easy to make a mistake that isn't caught by a human here
- One less thing to worry about during code reviews

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -67,21 +67,36 @@ file.
 
 ## Configure
 
-If possible, map the SDK API as the client used by Terraform.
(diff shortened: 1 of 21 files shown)


alrayyes:
> IMHO, this approach [#147 (comment)](https://github.com/Leaseweb/terraform-provider-leaseweb/pull/147#issuecomment-2467939139) could mitigate both of the issue. No more utils, no more copy the same thing over and over. Just implement once and use everywhere.

Create a PR, this is trivial to implement.

PR #158, opened 2024-11-15
feat: implement load balancer listener resource

alrayyes:


internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/utils"
+)
+
+var (
+	_ resource.ResourceWithConfigure   = &loadBalancerListenerResource{}
+	_ resource.ResourceWithImportState = &loadBalancerListenerResource{}
+)
+
+type loadBalancerListenerDefaultRuleResourceModel struct {
+	TargetGroupId types.String `tfsdk:"target_group_id"`
```
imtiazPabel, line 30: Nice to be `TargetGroupID`

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+		"chain":       types.StringType,
+	}
+}
+
+func (l loadBalancerListenerCertificateResourceModel) generateSslCertificate() publicCloud.SslCertificate {
+	sslCertificate := publicCloud.NewSslCertificate(
+		l.PrivateKey.ValueString(),
+		l.Certificate.ValueString(),
+	)
+	if !l.Chain.IsNull() && l.Chain.ValueString() != "" {
```
imtiazPabel, line 68: Since we've used the abstraction lots of time, would you like to use the method `AdaptStringPointerValueToNullableString` in here?

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+	Port           types.Int32  `tfsdk:"port"`
+	Certificate    types.Object `tfsdk:"certificate"`
+	DefaultRule    types.Object `tfsdk:"default_rule"`
+}
+
+func (l LoadBalancerListenerResourceModel) generateLoadBalancerListenerCreateOpts(ctx context.Context) (
+	*publicCloud.LoadBalancerListenerCreateOpts,
+	error,
+) {
+	protocol, err := publicCloud.NewProtocolFromValue(l.Protocol.ValueString())
```
imtiazPabel, line 102: Does this syntax seem more straightforward `publicCloud.Protocol(l.Protocol.ValueString())`?
No need to bother about `err`

internal/provider/publiccloud/load_balancer_listener_resource.go
alrayyes: That won't work. Here we explicitly only want to set the chain if it's not an empty string.

internal/provider/publiccloud/load_balancer_listener_resource.go
alrayyes: Didn't we agree to use get/set options where possible? Here the code should handle an error if an invalid string is passed.

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("generateLoadBalancerListenerCreateOpts: %v", err)
+	}
+
+	defaultRule := loadBalancerListenerDefaultRuleResourceModel{}
+	defaultRuleDiags := l.DefaultRule.As(ctx, &defaultRule, basetypes.ObjectAsOptions{})
+	if defaultRuleDiags != nil {
+		return nil, utils.ReturnError("generateLoadBalancerListenerCreateOpts", defaultRuleDiags)
```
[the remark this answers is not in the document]
alrayyes: Yes, that is handled in `create`. For other logic we do want to return errors the "normal" way, vs passing & setting response.Diagnostics every time.

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+
+	if !l.Protocol.IsNull() {
+		protocol, err := publicCloud.NewProtocolFromValue(l.Protocol.ValueString())
+		if err != nil {
+			return nil, fmt.Errorf("generateLoadBalancerListenerUpdateOpts: %v", err)
+		}
+		opts.SetProtocol(*protocol)
+	}
+
+	if !l.Port.IsNull() {
```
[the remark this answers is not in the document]
alrayyes: Port is required & can be null https://developer.leaseweb.com/api-docs/publiccloud_v1.html#tag/Load-Balancer/operation/updateLoadBalancerListener

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+	response *resource.ReadResponse,
+) {
+	var requestState LoadBalancerListenerResourceModel
+	response.Diagnostics.Append(request.State.Get(ctx, &requestState)...)
+	if response.Diagnostics.HasError() {
+		return
+	}
+
+	summary := fmt.Sprintf(
+		"Reading data %s for load_balancer_id %q listener_id %q",
```
imtiazPabel, line 421: typo `Reading resource`

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+	_ resource.SchemaRequest,
+	response *resource.SchemaResponse,
+) {
+	response.Schema = schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"load_balancer_id": schema.StringAttribute{
+				Required:    true,
+				Description: "Load balancer ID",
+			},
+			"listener_id": schema.StringAttribute{
```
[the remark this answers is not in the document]
alrayyes: i'm going with the api definition here. The whole schema is a mess.

internal/provider/publiccloud/load_balancer_listener_resource.go
```diff
@@ -0,0 +1,568 @@
...
+		utils.Error(ctx, &response.Diagnostics, summary, err, httpResponse)
+	}
+}
+
+func (l *loadBalancerListenerResource) ImportState(
+	ctx context.Context,
+	request resource.ImportStateRequest,
+	response *resource.ImportStateResponse,
+) {
+	idParts := strings.Split(request.ID, ",")
```
imtiazPabel, line 539: I'm curious to know, is this approach terraform suggests or there doesn't have any instruction from terraform, you came up with it yourself.

internal/provider/publiccloud/load_balancer_listener_resource.go
imtiazPabel: My experience says, in that case, we should also check `unknown`.  Otherwise, any edge case could come and set the value of `unknown`.

internal/provider/publiccloud/load_balancer_listener_resource.go
alrayyes: https://developer.hashicorp.com/terraform/plugin/framework/resources/import

internal/provider/publiccloud/load_balancer_listener_resource.go
imtiazPabel: It's not gonna possible to pass an invalid string, validation will protect it. I just tried to offer you a more convenient way, but if you found this is convenient to you, I'm fine also.

internal/provider/publiccloud/load_balancer_listener_resource.go
alrayyes: Usually i also add unknown, but that's never an issue here. Chain is either a string or null.

internal/provider/publiccloud/load_balancer_listener_resource.go
alrayyes: That's true. The validation should catch that. Even if the validation doesn't the server will. I'll remove it.

internal/provider/publiccloud/load_balancer_listener_resource.go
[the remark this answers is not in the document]
imtiazPabel: You are right, handle in `create`. Anything way you can decide, but just to ensure, maintaining the same way.

internal/provider/publiccloud/load_balancer_listener_resource.go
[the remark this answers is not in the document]
imtiazPabel: But we made it required in schema and there is a validation as well. No chance to come as `Null` in terraform IMHO

internal/provider/publiccloud/load_balancer_listener_resource.go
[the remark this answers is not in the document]
alrayyes: This is the golang way. As far as I understand it errors should bubble up. It's the providers responsibility to show/not show whatever to the end user. Keep things as simple as possible.

internal/provider/publiccloud/load_balancer_listener_resource.go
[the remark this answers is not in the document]
alrayyes: When updating it call be null. I think that happens when an empty string is passed, but i will double check.

internal/provider/publiccloud/load_balancer_listener_resource.go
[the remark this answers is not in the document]
imtiazPabel: For example, api schema says `instanceId`, but we specify as `id` in terraform schema.
https://developer.leaseweb.com/api-docs/publiccloud_v1.html#tag/Instances/operation/updateInstance

That's same for the other resources also.

internal/provider/publiccloud/load_balancer_listener_resource.go
imtiazPabel: TIL 😊

internal/provider/publiccloud/load_balancer_listener_resource.go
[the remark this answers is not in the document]
alrayyes: I know, I wrote that. The problem here is that the key is load_balancer_id & listener_id. I have no idea why, but that's the way it is in the api. That's why I chose listener_id and not id.

PR #161, opened 2024-11-18
feat(docs): let end user know which actions aren't supporetd

alrayyes:


internal/utils/schema_helpers.go
```diff
@@ -0,0 +1,64 @@
...
+type Action int
+
+const (
+	CreateAction Action = iota
+	ReadAction
+	UpdateAction
+	DeleteAction
+)
+
+func (a Action) String(unsupportedActions []Action) string {
```
imtiazPabel, line 20: I found the method `String`, isn't exported.

internal/provider/publiccloud/image_resource.go
```diff
@@ -149,11 +149,10 @@ func (i *imageResource) Metadata(
 
 func (i *imageResource) Schema(
 	_ context.Context,
-	_ resource.SchemaRequest,
+	request resource.SchemaRequest,
```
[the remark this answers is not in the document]
alrayyes: That can probably be removed now. At some point I did use it for something.

internal/utils/schema_helpers.go
alrayyes: That is correct, will fix.

PR #166, opened 2024-11-20
feat: implement target group resource

alrayyes:


internal/provider/publiccloud/target_group_resource.go
```diff
@@ -0,0 +1,475 @@
...
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/provider/client"
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/utils"
+)
+
+var (
+	_ resource.ResourceWithConfigure   = &targetGroupResource{}
+	_ resource.ResourceWithImportState = &targetGroupResource{}
+)
+
+type TargetGroupResourceModel struct {
```
imtiazPabel, line 29: Wondering, is it something exported or a typo?

alrayyes: typo

internal/provider/publiccloud/target_group_resource.go
```diff
@@ -0,0 +1,475 @@
...
+	error,
+) {
+	opts := publicCloud.NewCreateTargetGroupOpts(
+		t.Name.ValueString(),
+		publicCloud.Protocol(t.Protocol.ValueString()),
+		t.Port.ValueInt32(),
+		publicCloud.RegionName(t.Region.ValueString()),
+	)
+
+	if !t.HealthCheck.IsNull() {
```
imtiazPabel, line 49: My instinct suggests, it might be better to check for `isUnknown` as well.

internal/provider/publiccloud/target_group_resource.go
```diff
@@ -0,0 +1,475 @@
...
+	sdkHealthCheck, _ := sdkTargetGroup.GetHealthCheckOk()
+
+	healthCheck, err := utils.AdaptNullableSdkModelToResourceObject(
+		sdkHealthCheck,
+		healthCheckResourceModel{}.attributeTypes(),
+		ctx,
+		adaptHealthCheckToHealthCheckResource,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("adaptInstanceToInstanceResource: %w", err)
```
imtiazPabel, line 112: Can this error also be accomade like `utils.ReturnError`?
Just wanna make sure, to follow a single way.

internal/provider/publiccloud/target_group_resource.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+	targetGroup.HealthCheck = healthCheck
+
+	return &targetGroup, nil
+}
+
+type healthCheckResourceModel struct {
+	Protocol types.String `tfsdk:"protocol"`
+	Method   types.String `tfsdk:"method"`
+	Uri      types.String `tfsdk:"uri"`
```
imtiazPabel, line 122: `Uri` is an acronym, might be better to specify as `URI`

internal/provider/publiccloud/target_group_resource.go
```diff
@@ -0,0 +1,475 @@
...
+}
+
+func (h healthCheckResourceModel) generateOpts() publicCloud.HealthCheckOpts {
+	opts := publicCloud.NewHealthCheckOpts(
+		publicCloud.Protocol(h.Protocol.ValueString()),
+		h.Uri.ValueString(),
+		h.Port.ValueInt32(),
+	)
+
+	if !h.Method.IsNull() {
```
imtiazPabel, line 144: https://github.com/Leaseweb/terraform-provider-leaseweb/pull/166/files#r1853584742

internal/provider/publiccloud/target_group_resource.go
alrayyes: I think this works because `heath_check` is optional. So if it's no set the nil is assumed. I've tested this manually without setting health_check, no problems at all.

internal/provider/publiccloud/target_group_resource.go
alrayyes: We discussed this before. The ReturnError is handled in Create/Update. Otherwise you have to pass diags to every single function.

internal/provider/publiccloud/target_group_resource.go
alrayyes: Same thing, this is an optional field.

internal/provider/publiccloud/target_group_resource.go
imtiazPabel: My experience says, there could be some edge case, might happen during update.
I personally doesn't feel comfortable, without also checking `unknown`.
But if you're confident and have tested enough, then good to go.

internal/provider/publiccloud/target_group_resource.go
```diff
@@ -0,0 +1,475 @@
...
+		utils.Error(ctx, &response.Diagnostics, summary, err, httpResponse)
+		return
+	}
+
+	targetGroup, resourceErr := adaptTargetGroupToTargetGroupResource(
+		*sdkTargetGroup,
+		ctx,
+	)
+	if resourceErr != nil {
+		response.Diagnostics.AddError(summary, utils.DefaultErrMsg)
```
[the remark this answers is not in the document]
alrayyes: you're right, I missed this!

internal/provider/publiccloud/target_group_resource.go
imtiazPabel: In that case, we should come up in a common ground.
You know, it doesn't seems good to trigger error in 2 different ways.
If anything have in your mind, you can implement, otherwise we might need a seperate PR/story for that.

alrayyes: Just keep it simple. In symfony you also don't expect dependencies to set the http response. It's the terraform providers responsibility to set diags for the user. Everything else should be done the "go" way.

imtiazPabel (approved):

majidkarimizadeh (approved):

PR #176, opened 2024-11-25
Clean up code

alrayyes:
- Remove pointless return statement
- Rename variables that reference config, state or plan


ARCHITECTURE.md
```diff
@@ -119,3 +119,6 @@ The test functions themselves must adhere to the
 Public Cloud instance resource should be called `TestAccPublicCloudInstanceResource` and a test for an
 Public Cloud instance data source should be called `TestAccPublicCloudInstanceDataSource`.
 
+## Variables
+
+Where possible use `plan`,`state` & `config` as names for variables that reference terraform plan, state & config.
```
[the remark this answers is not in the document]
alrayyes: you can only access config in data sources. There is no plan.

internal/provider/dedicatedserver/contract_panels_data_source.go
```diff
@@ -73,20 +73,19 @@ func (c *controlPanelsDataSource) Read(
 	req datasource.ReadRequest,
 	resp *datasource.ReadResponse,
 ) {
-
-	var data controlPanelsDataSourceModel
-	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
+	var config controlPanelsDataSourceModel
```
imtiazPabel, line 76: Since, we're renaming the variable. Would we like to follow as terraform documentation?

alrayyes: it's `req.Config.Get`, not `req.Plan.Get`. Config is the correct term.

alrayyes: ah, now i see.

alrayyes: See the next line

internal/provider/dedicatedserver/contract_panels_data_source.go
```diff
@@ -107,16 +106,15 @@ func (c *controlPanelsDataSource) Read(
...
-
-	diags := resp.State.Set(ctx, &newData)
-	resp.Diagnostics.Append(diags...)
-	if resp.Diagnostics.HasError() {
-		return
-	}
+	resp.Diagnostics.Append(
+		resp.State.Set(
+			ctx,
+			controlPanelsDataSourceModel{
```
imtiazPabel, line 112: Just a heads-up. If you've already taken it into account, then we're fine.
Before refactor, it was implemented as a `reference(&)`, which is not anymore.

alrayyes: This should work. Nothing is actually written to the state here. Acceptance tests all pass. The documentation also passes nothing by reference https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-data-source-read

ARCHITECTURE.md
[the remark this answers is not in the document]
imtiazPabel: I don't mean, it's to be `plan`. 
Rather I mean, consensus come from, Terraform documentation https://developer.hashicorp.com/terraform/tutorials/providers-plugin-framework/providers-plugin-framework-data-source-read#implement-read-functionality

Please have a look, how they specify in `data_source read method`.

internal/provider/dedicatedserver/contract_panels_data_source.go
imtiazPabel: Again, please take a look, how terraform is showing to us.

internal/provider/dedicatedserver/contract_panels_data_source.go
imtiazPabel: If it's already taken it into account, then fine. But TBH, I don't trust the acceptance test completely.

imtiazPabel (approved):

PR #188, opened 2024-11-29
feat: implement public cloud isos data source

arunvarghe:


internal/provider/publiccloud/isos_data_source.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (i *isosDataSource) Read(
+	ctx context.Context,
+	_ datasource.ReadRequest,
+	resp *datasource.ReadResponse,
+) {
+	isos, httpResponse, err := getIsos(ctx, i.client)
+	if err != nil {
+		summary := fmt.Sprintf("Reading data %s", i.name)
```
alrayyes, line 127: I would get rid of the summary variable as it adds nothing and use `fmt.Sprintf` statement itself as a parameter for `utils.Error`

alrayyes (changes requested):
Small nitpick

alrayyes (approved):

internal/provider/publiccloud/isos_data_source.go
```diff
@@ -0,0 +1,167 @@
...
+	_ datasource.DataSourceWithConfigure = &isosDataSource{}
+)
+
+type isoDataSourceModel struct {
+	ID   types.String `tfsdk:"id"`
+	Name types.String `tfsdk:"name"`
+}
+
+type isosDataSourceModel struct {
+	Isos []isoDataSourceModel `tfsdk:"isos"`
```
imtiazPabel, line 27: `iso` is an acronym.
In that case, should be all capital letter `ISO`, like `ID` which recommended by `GO`.
Same goes for the method name as well.

imtiazPabel:
I’m a bit skeptical about using the word `isos` as a data source name. 
I understand it’s likely used for pluralization, but I feel it would be worth quickly checking with the team before merging or releasing to see if they’re comfortable with it.

alrayyes:
> I’m a bit skeptical about using the word `isos` as a data source name. I understand it’s likely used for pluralization, but I feel it would be worth quickly checking with the team before merging or releasing to see if they’re comfortable with it.

What else do you want to call it? Naming convention is `ISO` in capitals. Multiple of ISO is ISOs. The whole reason we having naming conventions is to keep things simple.

imtiazPabel:
> > I’m a bit skeptical about using the word `isos` as a data source name. I understand it’s likely used for pluralization, but I feel it would be worth quickly checking with the team before merging or releasing to see if they’re comfortable with it.
> 
> What else do you want to call it? Naming convention is `ISO` in capitals. Multiple of ISO is ISOs. The whole reason we having naming conventions is to keep things simple.

You're right, by principal it's correct.
But sounds a bit weird to me (might be only to me) , so that, just recommended for a double-checking with other team member.

alrayyes:
> > > I’m a bit skeptical about using the word `isos` as a data source name. I understand it’s likely used for pluralization, but I feel it would be worth quickly checking with the team before merging or releasing to see if they’re comfortable with it.
> > 
> > 
> > What else do you want to call it? Naming convention is `ISO` in capitals. Multiple of ISO is ISOs. The whole reason we having naming conventions is to keep things simple.
> 
> You're right, by principal it's correct. But sounds a bit weird to me (might be only to me) , so that, just recommended for a double-checking with other team member.

Once you start using these acronyms things start to sound weird, that's just the way it is. A lot of variable naming in golang sounds weird.

internal/provider/publiccloud/isos_data_source.go
```diff
@@ -0,0 +1,167 @@
...
+		ID:   basetypes.NewStringValue(sdkISO.GetId()),
+		Name: basetypes.NewStringValue(sdkISO.GetName()),
+	}
+}
+
+func getISOs(
+	ctx context.Context,
+	api publiccloud.PubliccloudAPI,
+) ([]publiccloud.Iso, *http.Response, error) {
+	var ISOs []publiccloud.Iso
```
alrayyes, line 52: this should lowercase as `ISO` is never exported. So `isos`

alrayyes (changes requested):

PR #190, opened 2024-11-29
test: write acceptance tests for dedicated_server resource

alrayyes:


internal/provider/dedicatedserver/server_resource.go
```diff
@@ -456,10 +456,20 @@ func (s *serverResource) Update(
 }
 
 func (s *serverResource) Create(
-	_ context.Context,
+	ctx context.Context,
```
imtiazPabel, line 459: In this case, would you also apply the same thing for `Delete` method.

alrayyes: Here you want nothing to happen if a resource is removed from the config. Otherwise any resource you add to main.tf, you can never ever remove.

imtiazPabel (approved):

PR #192, opened 2024-11-29
fix(docs): fix grammar for resources/datasources that can only be imp…

alrayyes:
…orted and not created


internal/utils/schema_helpers_test.go
```diff
@@ -32,38 +32,57 @@ func TestAction_firstAction(t *testing.T) {
...
+		func(t *testing.T) {
+			tests := []struct {
+				name string
+				a    Action
+				want string
+			}{
+				{
+					name: "create action",
+					a:    CreateAction,
+					want: "Once created, this resource cannot be created",
```
imtiazPabel, line 46: I think, this is not right expectation.
`Once created, this resource cannot be created`

alrayyes: I don't return an error here because string() should always return a string. A check could be implemented here with a `log.Fatal`, but this code will never be reached IRL.

imtiazPabel: I mean, the word, is self-contradictory.
Once created then again cannot be created.

imtiazPabel: I'm not agreed on the last line. 
If we wanted to allow being silly, then we didn't had to check and run the `log.Fatal` https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/internal/utils/schema_helpers.go#L38

I also, bring it to your attention, what will happen if it goes to `log.Fatal`?
As our expectation, is not only to stop the program, also show the customer a readable (at least the default) message. I just wanted to know, did you check that case?

PR #203, opened 2024-12-06
feat: implement ISO public cloud resource

alrayyes:


alrayyes:
Will fix merge issues after approval

internal/provider/publiccloud/instance_iso_resource.go
```diff
@@ -0,0 +1,455 @@
...
+	"github.com/leaseweb/terraform-provider-leaseweb/internal/utils"
+)
+
+var (
+	_ resource.ResourceWithConfigure   = &instanceISOResource{}
+	_ resource.ResourceWithImportState = &instanceISOResource{}
+)
+
+type invalidISOIDError struct {
+	supportedISOIds []string
```
imtiazPabel, line 25: `supportedISOIDs`

internal/provider/publiccloud/instance_resource.go
```diff
@@ -58,6 +77,7 @@ type instanceResourceModel struct {
 	Region              types.String `tfsdk:"region"`
 	Reference           types.String `tfsdk:"reference"`
 	Image               types.Object `tfsdk:"image"`
+	Iso                 types.Object `tfsdk:"iso"`
```
imtiazPabel, line 80: `ISO`

docs/resources/public_cloud_instance.md
```diff
@@ -185,6 +186,15 @@ Read-Only:
 
 - `ip` (String)
 
+
+<a id="nestedatt--iso"></a>
+### Nested Schema for `iso`
+
+Read-Only:
+
+- `iso_id` (String) The ISO ID.
```
imtiazPabel, line 195: Can it be only `id`, as it is in the data_source https://github.com/Leaseweb/terraform-provider-leaseweb/blob/main/docs/data-sources/public_cloud_instances.md?plain=1#L95

docs/resources/public_cloud_instance_iso.md
```diff
@@ -0,0 +1,46 @@
...
+
+- `instance_id` (String) Instance's ID. This value cannot be changed.
+
+### Optional
+
+- `desired_iso_id` (String) The desired ISO ID. Removing this will detach the current ISO from the instance. Changing it will cause the current ISO to be detached and a new one to be attached to the instance.
+
+### Read-Only
+
+- `iso_id` (String) The ISO ID. Detaching/attaching iso_ids is an asynchronous operation. This attribute shows the current iso_id while `desired_iso_id` has the desired iso_id state.
```
imtiazPabel, line 36: Can it also be `id`?

internal/provider/publiccloud/instance_iso_resource.go
```diff
@@ -0,0 +1,455 @@
...
+	response *resource.SchemaResponse,
+) {
+	response.Schema = schema.Schema{
+		Description: "Creating this resource attaches the ISO to the instance, deleting it detaches the ISO from the instance.",
+		Attributes: map[string]schema.Attribute{
+			"desired_iso_id": schema.StringAttribute{
+				Optional:    true,
+				Description: "The desired ISO ID. Removing this will detach the current ISO from the instance. Changing it will cause the current ISO to be detached and a new one to be attached to the instance.",
+			},
+			"instance_id": schema.StringAttribute{
```
imtiazPabel, line 107: Since we're expecting not to change `instance_id`, event if it's change, it should detach old one and attach with new one. In that case, `RequireReplace` might be a good choice.

internal/provider/publiccloud/instance_iso_resource.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+func NewInstanceIsoResource() resource.Resource {
+	return &instanceISOResource{
+		name: "public_cloud_instance_iso",
+	}
+}
+
+// updateISO detaches the current ISO if there is anything attached. If api.desired_api_id is set then a new ISO is attached.
+func updateISO(
```
imtiazPabel, line 342: `updateISO` performs both `attach/detach` depends on the circumstances, which is a bit overwhelming. I'd appreciate, if these could be split.

alrayyes: I tried that, the code becomes even more hard to read. Whether attaching or detaching, you always have to check of an iso is attached currently and take that into account.

Splitting this function up adds little as you end up with a detachIfISOAttwched function that's untestable and only called once.

There is a reason to the madness.

internal/provider/publiccloud/instance_iso_resource.go
alrayyes: This will confuse the user. Changing the instance_id in this cased actually changes the state of 2 resources, which you don't want

docs/resources/public_cloud_instance_iso.md
alrayyes: This is confusing, as iso-id and instance_id is the joint key

arunvarghe: I see that we follow "id" in other places like loadbalancer and image etc.

docs/resources/public_cloud_instance.md
alrayyes: That doesn't make sense in this context. If you make it `id`, then `id` of what? With `iso_id` it's clear what we're talking about and it's absolutely clear to the end user what this is about.

docs/resources/public_cloud_instance_iso.md
alrayyes: Those are real resources that correlate to an endpoint, this does not.

docs/resources/public_cloud_instance.md
imtiazPabel: I'm wondering, if it doesn't make sense for resource, then how could it makes sense for the same data source.

Also, when a `id` comes under the `iso` object, it clearly understands that, it's mean a `iso id`. And this is the same way, you put `id` inside `image` object. Then why not for `ISO`

docs/resources/public_cloud_instance_iso.md
imtiazPabel: `instance_id & desired_iso_id` is the joint key I can see on the `ImportState` method.
If i'm correct, then I think, it could be `id` instead of `iso_id`

imtiazPabel (approved):

arunvarghe (approved):

majidkarimizadeh (approved):

PR #221, opened 2024-12-13
refactor: unify error handling

alrayyes:
According to the [documentation](https://developer.hashicorp.com/terraform/plugin/framework/diagnostics) helpers should be used to generate consistent diagnostics

> Create a helper function in your provider code using the diagnostic creation functions available in the [diag package](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/diag) to generate consistent diagnostics for types of errors/warnings. It is also possible to use [custom diagnostics types](https://developer.hashicorp.com/terraform/plugin/framework/diagnostics#custom-diagnostics-types) to accomplish this same goal.

As Terraform itself already tells us what resource/data source an error applies to the summary we use is redundant

2024-12-13T11:18:55,665836016+01:00

This PR adds helpers to facilitate all edge cases terraform currently & unify the error handling.

### DIFF
--- a/internal/provider/dedicatedserver/contract_panels_data_source.go
+++ b/internal/provider/dedicatedserver/contract_panels_data_source.go
@@ -46,14 +46,7 @@ func (c *controlPanelsDataSource) Configure(
 	coreClient, ok := req.ProviderData.(client.Client)
 
 	if !ok {
-		resp.Diagnostics.AddError(
-			"Unexpected Resource Configure Type",
-			fmt.Sprintf(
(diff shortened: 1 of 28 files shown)


imtiazPabel (approved):

imtiazPabel:
Since this is a significant change to the codebase(`error logging`). 
I would like to, bring it to the team's attention. Please raise any concerns (if have), we can still address them in next PR.
@majidkarimizadeh @alrayyes mostly

PR #313, opened 2025-02-17
Feat/dedicated server installation import

### LINKED ISSUE #300: Migration of server installation resources from 0.4.1
I have tried to upgrade our Terraform workspaces from 0.4.1 to 1.24.0, and the schema for `leaseweb_dedicated_server_installation` has changed in a way that prevents Terraform from being able to read the existing state and from deleting existing resources from the state:

```
Failed generating plan JSON
Exit code: 1

Failed to marshal plan to json: error marshaling prior state: unsupported attribute "job_uuid"
```

and

```
Error: Unable to Read Previously Saved State for UpgradeResourceState
with module.gc-install["gc02"].leaseweb_dedicated_server_installation.main
on .terraform/modules/gc-install/leaseweb/server/installation/main.tf line 7, in resource "leaseweb_dedicated_server_installation" "main":
resource "leaseweb_dedicated_server_installation" "main" {
There was an error reading the saved resource state using the current resource schema.

If this resource state was last refreshed with Terraform CLI 0.11 and earlier, it must be refreshed or applied with an older provider version first. If you manually modified the resource state, you will need to manually modify it to match the current resource schema. Otherwise, please report this to the provider developer:

AttributeName("raid"): invalid JSON, expected "{", got "["
```

I have mitigated this issue by

- removing all `leaseweb_dedicated_server_installation` resources from the state with the old provider version
- manually migrated the raw JSON snippets from the old schema to the new schema
- added the JSON snippets back to the state file
- manually pushed the state back to Terraform Cloud

This works, but it is pretty cumbersome. Instead of adding support for migrating from the old schema, would it be possible to add a "fake" import for `leaseweb_dedicated_server_installation` so that existing servers can be added and imported to Terraform code without forcing a reinstallation or manual intervention in the state file?

imtiazPabel:
Fixes #300


internal/provider/dedicatedserver/installation_resource.go
```diff
@@ -326,94 +331,72 @@ func (i *installationResource) Create(
...
 	}
 
-	resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
-}
+	serverID := state.DedicatedServerID.ValueString()
 
-func getJobStatus(serverID string, jobID string, i *installationResource, ctx context.Context, resp *resource.CreateResponse) (string, error) {
-	request := i.DedicatedserverAPI.GetJob(ctx, serverID, jobID)
+	result, response, err := i.DedicatedserverAPI.GetJobList(ctx, serverID).
+		Offset(0).Limit(1).Type_("install").Status("FINISHED").Execute()
```
majidkarimizadeh, line 379: is there a reason that we fetch only FINISHED jobs?

docs/resources/dedicated_server_installation.md
```diff
@@ -85,3 +85,12 @@ Optional:
   - *HW*
   - *SW*
   - *NONE*
+
+## Import
+
+Import is supported using the following syntax:
+
+```shell
+# Dedicated srever installed job can be imported by specifying the server id.
```
majidkarimizadeh, line 94: would be good to let customers know that we fetch only first job, otherwise I'd expect this command to get 2 params which is for serverid/jobid

imtiazPabel: Basically, having two parameters here isn’t realistic and introduces a breaking change with legacy.

The reason it’s not realistic is that a customer cannot obtain an installation `job ID` from the customer portal. The only way to retrieve it is by calling the list job `API` with `type` and `status` (what basically we do in codebase), then manually filtering the correct one, which would be very cumbersome for our customers.

If you, take a look the issue, which raised by the customer #300 and their exact pain point, passing a `job ID` is not a viable solution—it adds complexity rather than solving the problem.

Regarding informing the customer that we fetch only the latest job, I believe this is an internal abstraction. How we determine the job to `import` should remain our implementation detail, and it makes sense that the last job should be used. Explicitly informing the customer about this might not be necessary.

majidkarimizadeh (approved):

PR #375, opened 2025-06-11
feat: add instance to private network

dgomespt:
Allows adding an instance to a private network at creation and adding/removing an existing instance to a private network

### DIFF
--- a/docs/data-sources/public_cloud_instances.md
+++ b/docs/data-sources/public_cloud_instances.md
@@ -30,7 +30,7 @@ data "leaseweb_public_cloud_instances" "all" {}
 Read-Only:
 
 - `contract` (Attributes) (see [below for nested schema](#nestedatt--instances--contract))
-- `has_private_network` (Boolean) The instance should be added to a private network
+- `has_private_network` (Boolean) Indicates whether the instance is connected to a private network
 - `id` (String) The instance unique identifier
 - `image` (Attributes) (see [below for nested schema](#nestedatt--instances--image))
 - `ips` (Attributes List) (see [below for nested schema](#nestedatt--instances--ips))
--- a/docs/data-sources/public_cloud_instances.md
+++ b/docs/data-sources/public_cloud_instances.md
@@ -30,6 +30,7 @@ data "leaseweb_public_cloud_instances" "all" {}
 Read-Only:
 
 - `contract` (Attributes) (see [below for nested schema](#nestedatt--instances--contract))
+- `has_private_network` (Boolean) The instance should be added to a private network
 - `id` (String) The instance unique identifier
 - `image` (Attributes) (see [below for nested schema](#nestedatt--instances--image))
 - `ips` (Attributes List) (see [below for nested schema](#nestedatt--instances--ips))
--- a/internal/provider/publiccloud/instance_resource.go
+++ b/internal/provider/publiccloud/instance_resource.go
@@ -717,7 +717,7 @@ func (i *instanceResource) Schema(
 			"private_network": schema.BoolAttribute{
 				Computed:    true,
 				Optional:    true,
-				Description: "",
+				Description: "Indicates whether the instance is connected to a private network",
 			},
 		},
 	}
--- a/internal/provider/publiccloud/instance_resource.go
+++ b/internal/provider/publiccloud/instance_resource.go
@@ -75,7 +75,7 @@ type instanceResourceModel struct {
 	IPs                 types.List   `tfsdk:"ips"`
 	Contract            types.Object `tfsdk:"contract"`
 	MarketAppID         types.String `tfsdk:"market_app_id"`
-	HasPrivateNetwork   types.Bool   `tfsdk:"private_network"`
+	HasPrivateNetwork   types.Bool   `tfsdk:"has_private_network"`
 }
 
 func adaptInstanceDetailsToInstanceResource(
@@ -705,7 +705,7 @@ func (i *instanceResource) Schema(
 					stringplanmodifier.RequiresReplaceIfConfigured(),
 				},
 			},
-			"private_network": schema.BoolAttribute{
+			"has_private_network": schema.BoolAttribute{
 				Computed:    true,
 				Optional:    true,
 				Description: "Indicates whether the instance is connected to a private network",
--- a/internal/provider/publiccloud/instance_resource.go
+++ b/internal/provider/publiccloud/instance_resource.go
@@ -2,8 +2,11 @@ package publiccloud
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"time"
 
+	"github.com/cenkalti/backoff/v5"
 	"github.com/hashicorp/terraform-plugin-framework-validators/int32validator"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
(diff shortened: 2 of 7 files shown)


internal/provider/publiccloud/instance_resource.go
```diff
@@ -323,6 +346,135 @@ func (i *instanceResource) Read(
...
+	if plan.HasPrivateNetwork.ValueBool() && !instanceDetails.HasPrivateNetwork {
+
+		res, err = i.PubliccloudAPI.AddToPrivateNetwork(ctx, instanceDetails.Id).Execute()
+
+		if err != nil {
+			utils.SdkError(ctx, &resp.Diagnostics, err, res)
+			return
+		}
+
+		updated, err = i.waitUntilPrivateNetworkChanges(ctx, resp, instanceDetails.Id, true)
```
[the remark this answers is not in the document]
majidkarimizadeh: now I get it, turning off/on it's async again.

docs/data-sources/public_cloud_instances.md
```diff
@@ -30,6 +30,7 @@ data "leaseweb_public_cloud_instances" "all" {}
 Read-Only:
 
 - `contract` (Attributes) (see [below for nested schema](#nestedatt--instances--contract))
+- `has_private_network` (Boolean) The instance should be added to a private network
```
imtiazPabel, line 33: The description could be more clearer. f.e: `Indicates whether the instance is connected to a private network` or anything which is make more sense.

internal/provider/publiccloud/instance_resource.go
```diff
@@ -233,26 +239,42 @@ func (i *instanceResource) Create(
...
+			if plan.HasPrivateNetwork.ValueBool() {
+
+				res, err := i.PubliccloudAPI.AddToPrivateNetwork(ctx, changed.Id).Execute()
+
+				if err != nil {
+					utils.SdkError(ctx, &resp.Diagnostics, err, res)
+					return
+				}
+
+				changed.HasPrivateNetwork = true
```
imtiazPabel, line 257: I believe we should also `waitUntilPrivateNetworkChanges` over here, as we do during update process. Therefore, it would be more certain and we don't have to explicitly update the `HasPrivateNetwork` parameter, again the same way it's happening for the update.

internal/provider/publiccloud/instance_resource.go
```diff
@@ -323,6 +346,135 @@ func (i *instanceResource) Read(
...
+	plan instanceResourceModel,
+	instanceDetails *publiccloud.InstanceDetails,
+	ctx context.Context,
+	resp *resource.UpdateResponse,
+) {
+
+	var res *http.Response
+	var updated *publiccloud.InstanceDetails
+	var err error
+
```
imtiazPabel, line 359: My understanding says, the value could also be `unknown`.
However, what if in the beginning, we do a check like below.

```
if plan.HasPrivateNetwork.IsUnknown(){
return 
}
```

internal/provider/publiccloud/instance_resource.go
```diff
@@ -233,26 +239,42 @@ func (i *instanceResource) Create(
...
-		*instanceDetails,
-		ctx,
-		&resp.Diagnostics,
-	)
-	if resp.Diagnostics.HasError() {
+	if changed != nil {
+
+		if !plan.HasPrivateNetwork.IsUnknown() {
+
+			if plan.HasPrivateNetwork.ValueBool() {
```
imtiazPabel, line 248: We can also write both condition together, if we want to `if !plan.HasPrivateNetwork.IsUnknown() && plan.HasPrivateNetwork.ValueBool()`

internal/provider/publiccloud/instance_resource.go
```diff
@@ -323,6 +346,135 @@ func (i *instanceResource) Read(
...
+		}
+
+		// Sleep for the backoff interval before retrying
+		time.Sleep(bo.NextBackOff())
+		retryCount++
+	}
+
+}
+
+func (i *instanceResource) waitUntilPrivateNetworkChanges(
```
imtiazPabel, line 440: The implementation of both `waitUntilStateChanges & waitUntilPrivateNetworkChanges` are identical. It would be a good idea, to maintain a single method and depends on the parameter it could serve our purpose?

dgomespt: This is done in latest commit

internal/provider/publiccloud/instance_resource.go
```diff
@@ -323,6 +346,135 @@ func (i *instanceResource) Read(
...
+func (i *instanceResource) waitUntilStateChanges(
+	ctx context.Context,
+	resp *resource.CreateResponse,
+	instanceId string,
+	expectedState string,
+) (*publiccloud.InstanceDetails, error) {
+	// Create a constant backoff with a 5-second retry interval
+	bo := backoff.NewConstantBackOff(5 * time.Second)
+
+	// Set the retry limit to 10 retries (60 seconds)
```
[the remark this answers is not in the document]
dgomespt: This is addressed in latest commit.

internal/provider/publiccloud/instance_resource.go
```diff
@@ -379,93 +394,66 @@ func (i *instanceResource) TogglePrivateNetwork(
...
 		*instanceDetails = *updated
 
 	}
 
+	return nil, nil
+
 }
 
-func (i *instanceResource) waitUntilStateChanges(
+func (i *instanceResource) waitUntilPropertyChanges(
```
imtiazPabel, line 417: A small nitpicks
Since we're not only wait for the property change but also match with value. 
IMHO, `waitUntilPropertyMatch/WaitUntilPropertyEquals` or any other name which make more sense

internal/provider/publiccloud/instance_resource.go
```diff
@@ -239,28 +239,54 @@ func (i *instanceResource) Create(
...
+		if err != nil {
+			utils.SdkError(ctx, &resp.Diagnostics, err, res)
+			return
+		}
 
-				if err != nil {
-					utils.SdkError(ctx, &resp.Diagnostics, err, res)
-					return
-				}
+		state := adaptInstanceDetailsToInstanceResource(
```
imtiazPabel, line 267:
```
state := adaptInstanceDetailsToInstanceResource(
			*instanceDetails,
			ctx,
			&resp.Diagnostics,
		)
		if resp.Diagnostics.HasError() {
			return
		}

		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```

That common code block contains both in `if/else`. I think we can pass it at the end of the block and it should still work.

internal/provider/publiccloud/instance_resource.go
```diff
@@ -233,26 +239,42 @@ func (i *instanceResource) Create(
...
+			*changed,
+			ctx,
+			&resp.Diagnostics,
+		)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
+		resp.Diagnostics.Append(resp.State.Set(ctx, state)...)
```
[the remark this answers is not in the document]
dgomespt: From what I could find, terraform will update the sate at the end of Update() method regardless of how many times you set the resp state during execution.

imtiazPabel (approved):

